PR #1, opened 2022-09-14
Add project folders framework

JoeZiminski:


manager/TODO.py
```diff
@@ -0,0 +1,23 @@
+
```
adamltyson, line 1: I'd keep the TODO list out of the code, ideally this should be in GH issues, so that others can be assigned etc.

manager/manager.py
```diff
@@ -0,0 +1,1105 @@
+import copy
```
adamltyson, line 1: Nitpicking, but `isort` is very useful to organise imports by type. Maybe we should adopt the [brainreg pre-commit config](https://github.com/brainglobe/brainreg/blob/master/.pre-commit-config.yaml) to autoformat team projects?

manager/manager.py
```diff
@@ -0,0 +1,1105 @@
...
+
+# --------------------------------------------------------------------------------------------------------------------
+# Project Manager Class
+# --------------------------------------------------------------------------------------------------------------------
+
+
+class ProjectManager:
+    """
+    Main project manager class for data organisation and transfer in BIDS-style project directory.
+    The expected organisation is a central repository on a linux remote cluster ('remote') that
```
adamltyson, line 67: Will this only ever work with a Linux remote? Just thinking of long-term generalisation.

manager/manager.py
```diff
@@ -0,0 +1,1105 @@
...
+    each contain a subset of the full project (e.g. machine for electrophysiology collection,
+    machine for behavioural connection, machine for analysis for specific data etc.).
+
+    On first use on a new profile, the user will be prompted to set configurations with the function
+    make_config_file().
+
+    For transferring data between a remote data storage with SSH, use setup setup_ssh_connection_to_remote_server().
+    This will allow you to check the server Key, add host key to profile if accepted, and setup ssh key pair.
+
+    INPUTS: username - the username for the profile to use. Profile files are stored in your Appdir folder
```
adamltyson, line 78: It may be worth specifying for the user what username is this. It's not immediately clear whether this is for the local machine, the remote server, or maybe even for this software itself.

manager/manager.py
```diff
@@ -0,0 +1,1105 @@
...
+                },
+            ),
+            "behav": Folder(
+                "behav",
+                self.cfg["use_behav"],
+                subfolders={
+                    "behav_camera": Folder("camera", self.cfg["use_behav_camera"]),
+                },
+            ),
+            "microscopy": Folder(
```
adamltyson, line 136: As discussed, `microscopy` should be split into something like `imaging` (meaning e.g. 2p/widefield) and `histology`.

manager/manager.py
```diff
@@ -0,0 +1,1105 @@
...
+        sub_names: Union[str, list],
+        ses_names: Union[str, list],
+        preview: bool = False,
+    ):
+        """
+        Upload data from a local machine to the remote project folder.
+        In the case that a file / folder exists on the remote and local, the local will
+        not be overwritten even if the remote file is an older version.
+
+        TODO: this is highly sub-optimal performance wise when downloading using the "all" option
```
adamltyson, line 212: As above, I'd move TODO's to issues.

manager/manager.py
```diff
@@ -0,0 +1,1105 @@
...
+        :param sub_names: a list of sub names as accepted in make_sub_folder(). "all" will search for all
+                          sub- folders in the data type folder to upload.
+        :param ses_names: a list of ses names as accepted in make_sub_folder(). "all" will search each
+                          sub- folder for ses- folders and upload all.
+        :param preview: perform a dry-run of upload, to see which files are moved.
+        """
+        self._transfer_sub_ses_data("upload", data_type, sub_names, ses_names, preview)
+
+    def download_data(
+        self, data_type: str, sub_names: Union[str, list], ses_names: Union[str, list]
```
adamltyson, line 227: `data_type` could be confusing, this could mean, e.g. `pathlib.Path`, `.tiff`, or `ephys`. Maybe something like `experiment_type` would be clearer?

manager/manager.py
```diff
@@ -0,0 +1,1105 @@
...
+
+                        if make_ses_tree:
+                            self._make_ses_directory_tree(sub, ses, data_type_key)
+
+    def _make_ses_directory_tree(self, sub: str, ses: str, data_type_key: str):
+        """
+        Make the directory tree within a session. This is dependent on the data_type (e.g. "ephys")
+        folder and defined in the subfolders field on the Folder class, in self._ses_folders.
+
+        All subfolders will be make recursively, unless the .used attribute on the Folder class is
```
adamltyson, line 485: will be ~~make~~ made

manager/manager.py
```diff
@@ -0,0 +1,1105 @@
...
+        data_type_dir = self._ses_folders[data_type_key]
+
+        if data_type_dir.used and data_type_dir.subfolders:
+            self._recursive_make_subfolders(
+                folder=data_type_dir, path_to_folder=[data_type_dir.name, sub, ses]
+            )
+
+    def _recursive_make_subfolders(self, folder: Folder, path_to_folder: list):
+        """
+        Function to recursively create all directories in a Folder .subfolders field.
```
adamltyson, line 501: Nitpicking, but I'd use directory throughout, rather than swapping between folder/directory.

manager/manager.py
```diff
@@ -0,0 +1,1105 @@
...
+        opts = {
+            "help": False,
+            "verbose": 5,
+            "quiet": 0,
+            "debug ": False,
+            "case": "strict",
+            "dry_run": preview,
+            "progress": False,
+            "no_color": True,
+            "ftp_active": False,
```
adamltyson, line 663: Will these ever need to be changed, i.e. should they be hardcoded?

manager/manager.py
```diff
@@ -0,0 +1,1105 @@
...
+        :param config_dict: self.cfg dict of configs
+        :param direction: "path_to_str" or "str_to_path"
+
+        TODO: put "local_path" and "remote_path" keys somewhere central.
+        """
+        for path_key in ["local_path", "remote_path"]:
+            if direction == "str_to_path":
+                config_dict[path_key] = Path(config_dict[path_key])
+            elif direction == "path_to_str":
+                config_dict[path_key] = config_dict[path_key].as_posix()
```
adamltyson, line 946: Out of curiosity, is `as_posix()` preferred to `str()`?

adamltyson (changes requested):
I haven't gone through in too much detail, but all looks good. I've left a few specific comments. More generally, it would be useful to start adding docs and tests now so it's easier for other to test it as it develops. 

`manager.py` is also a very long file, and hard to review. I'd move as much code out into other modules as possible. 

Lastly, the code line length is formatted, but it would be good to do the same with the docstrings. Some are very long.

manager/manager.py
JoeZiminski: Thanks good idea I have added as pre-commit, I did run isort on it but maybe made changes following - thus good to have pre-commit!

manager/manager.py
JoeZiminski: I think it should generalize easily, in fact if remote is mounted it does not require linux. The docstring is confusing because the only linux requirement is around writing the SSH key to authorized users, I will remove it and add it as a note on a TODO / issue

manager/manager.py
JoeZiminski: edited username was not a great description, project_name more appropriate as for each project a new config must be setup (including root dir tree etc).

manager/manager.py
JoeZiminski: I'm not certain, currently they are set to be as un-intrusive as possible (e.g. never overwrite an existing file). In theory a user might want to edit them, and during this testing phase we might make amendments. But generally, there will probably be a default set that is strongly advised to use.

manager/manager.py
JoeZiminski: hmm good question, I used posix() with some vague notion to keep all outputs to forward slashes OS system independent, but not sure if it actually serves any purpose. 

It seems os.fspath() [is preferred](https://stackoverflow.com/questions/56477425/real-path-of-file-string-pathlib), shall default to this?

manager/manager.py
JoeZiminski: thanks agree much nicer

manager/manager.py
adamltyson: Sounds good.

JoeZiminski: In the end I used posix() for 'internal' functions as was running into an issue in which sections of paths generated from the windows local project were not compatible with unix system. However for public functions (e.g. project.get_local_path()) the os.fspath is used.

manager/manager.py
JoeZiminski: yes I think so, and it would be nice to have at least a subset of these as possible options. I'm testing at the moment and will check if the pyftpsync provides functionality to deal with disrupted connections etc. If not, it is probably worth jettisoning this and using rsync / robocopy depending on os.

JoeZiminski:
> `manager.py` is also a very long file, and hard to review. I'd move as much code out into other modules as possible.

agree, have factored out configs and some utils. ATM utils of different type (ssh, directory managerment) are combined into a utils file but will be worth factoring out of separate files as it gros.

> Lastly, the code line length is formatted, but it would be good to do the same with the docstrings. Some are very long.

thanks have done. 


Am currently writing tests which will be useful to see how optimal pyftpsync is for our needs. It might be worth using rysnc / robocopy directly rather than pyftpsync which uses sftp commands directly, but doesn't provide as much extra checking (e.g. checksum options).

adamltyson:
> Am currently writing tests which will be useful to see how optimal pyftpsync is for our needs. It might be worth using rysnc / robocopy directly rather than pyftpsync which uses sftp commands directly, but doesn't provide as much extra checking (e.g. checksum options).

Sounds good. I guess there's a package out there for doing these checks?

PR #9, opened 2022-10-20
Adding rclone

JoeZiminski:
PR to merge Rclone changes into adding_tests branch. 

Currently, RClone is only setup for windows, I'll extend to macos and linux next in the adding_tests branch as testing on these systems.

RClone is an executable file so it is downloaded from RClone website based on the user OS. It is downloaded to the user appdir. RClone is on conda, so maybe conda is a better way?

The download function from http_utils is a direct copy of the bgatlas function. Note sure of the best way to approach this. 

RClone does not copy empty directories, which led to inconsistent behaviour when copying session-level directories (i.e. the behav session level dir has folders in, but not imaging / histology. As such the latter were not copied in tests). I think we will want to write a meta-data file at some point in future, so include .datashuttle_meta/ dir in all session level dirs for now.


datashuttle/configs.py
```diff
@@ -18,42 +18,102 @@ def __init__(self, file_path, input_dict):
...
-                " provided"
+                " provided."
             )
 
     def update_an_entry(self, option_key: str, new_info: Union[str, bool]):
         """
         Convenience function to update individual entry of configuration file.
         The config file, and currently loaded self.cfg will be updated.
 
+        Note if user updates connection type, a different remote_path
```
adamltyson, line 68: General point, but I would write these docstrings *for* the user, rather than for developers, referencing the user. If we release this with a Python API, we would expect them to dig into the docstrings.

datashuttle/configs.py
```diff
@@ -18,42 +18,102 @@ def __init__(self, file_path, input_dict):
...
+            warnings.warn(f"{option_key} was not updated")
+            self[option_key] = original_value
+
+    def safe_check_current_dict_is_valid(self):
+        """ """
+        try:
+            self.check_dict_values_and_inform_user()
+            return True
+
+        except BaseException as e:  # TODO: test invalid change
```
adamltyson, line 114: Tiny point, but it would be great to have TODOs as GitHub issues, so that others can see at a glance what could be worked on.

datashuttle/datashuttle.py
```diff
@@ -281,11 +283,13 @@ def make_config_file(
         settings will be used each time the datashuttle is opened.
 
         :param local_path:                  path to project dir on local machine
-        :param remote_path:                 path to project directory on remote machine. Note this cannot
-                                            include ~ home directory syntax, must contain the full path (
-                                            e.g. /nfs/nhome/live/jziminski)
+        :param remote_path_local:           Full filepath to local filesystem (e.g. mounted drive) dir
```
adamltyson, line 286: Is there any way of ssh'ing to the remote machine to expand the full path, so we could support the `~` syntax?

adamltyson: Not critical, happy to enforce the use of full paths.

datashuttle/datashuttle.py
```diff
@@ -389,7 +400,71 @@ def get_appdir_path(self):
         return os.fspath(appdir_path)
 
     def get_remote_path(self):
-        return self.cfg["remote_path"].as_posix()
+        """
+        Force remote path to return as posix as if local fs is windows and remote is
```
adamltyson, line 404: I guess fs is filesystem? I'd expand these acronyms so it's easier to parse in a years time.

datashuttle/utils_mod/rclone_utils.py
```diff
@@ -0,0 +1,166 @@
...
+import os.path
+import shutil
+import subprocess
+import warnings
+from pathlib import Path
+
+from datashuttle.utils_mod import http_utils, utils
+
+
+def call_rclone(command : str, silent : bool = False):
```
adamltyson, line 11: Have you tested [this](https://github.com/rclone/rclone/blob/d2fef05fe45074c458ae11d6b3361a2bd26c9c10/librclone/python/rclone.py) rather than calling a subprocess?

datashuttle/tests/test_configs.py
```diff
@@ -48,11 +48,112 @@ def test_warning_on_startup(self):
...
+    def test_fail_to_pass_remote_path(self, project):
+        """
+        Test that the make_config_file will assert if neither
+        remote_path_ssh or remote_path_local are passed.
+        """
+        try:
+            project.make_config_file("test_local_path", False)
+            assert False, "expected error was not caught"
+
+        except AssertionError as e:
```
adamltyson, line 60: Would [`pytest.raises()`](https://docs.pytest.org/en/stable/reference/reference.html?highlight=raises#pytest.raises) be simpler for testing these errors?

datashuttle/utils_mod/http_utils.py
```diff
@@ -0,0 +1,66 @@
+# TODO: this is directly copy from https://github.com/brainglobe/bg-atlasapi/blob/5835887c5f9fac16ea1f2682f861411e4f789b3c/bg_atlasapi/utils.py#L132
+
+from typing import Callable
+
+import requests
+from rich.progress import (BarColumn, DownloadColumn, Progress, TextColumn,
+                           TimeRemainingColumn, TransferSpeedColumn)
+
+
+def retrieve_over_http(
```
adamltyson, line 10: As mentioned in the PR, I would vendor this clearly and include the relevant license.

adamltyson (changes requested):

adamltyson (approved):

adamltyson:
> RClone is an executable file so it is downloaded from RClone website based on the user OS. It is downloaded to the user appdir. RClone is on conda, so maybe conda is a better way?

Depending on the license, we could also include pre-built binaries, but I think just putting everything on conda is likely the way to go (with this and everything else).

> If it's just this function and not any more of the package, I'd say just vendor it. It's academic here as it's "our" code, but we should define a best practice for the team in these cases. I would usually include a function like this in the following way:

```
package/
├─ subdir1/
├─ subdir2/
├─ vendored/
  ├─ vendored_package1/
       ├─ LICENSE.txt
       ├─ vendored_code.py

```

> RClone does not copy empty directories, which led to inconsistent behaviour when copying session-level directories (i.e. the behav session level dir has folders in, but not imaging / histology. As such the latter were not copied in tests). I think we will want to write a meta-data file at some point in future, so include .datashuttle_meta/ dir in all session level dirs for now.

Seems reasonable. 

I've left a bunch of comments, but approved the PR anyway if you want to just get to merge and get to work on the testing branch.

datashuttle/datashuttle.py
JoeZiminski: I think it would be nice and if I recall correctly was based on limitation of pyftpsync but rclone (99% sure) does support ~ syntax, will look into before merging into master #10

datashuttle/utils_mod/rclone_utils.py
JoeZiminski: unfortunately leads to some problems with compilation across systems ([discussed here](https://forum.rclone.org/t/platform-independent-use-of-librclone-python-api/33286) - the rlcone people are very helpful!).

However, this is a API that spins a HTTP server to use. This is nice as it means the job can be communicated to from other machines. However I don't know much about HTTP servers and was worried about it creating problems (vague, unspecified problems so would probably be okay) , so have avoided for now with an eye to extend in future. However, do you think it might be worth using this from the outset?

datashuttle/configs.py
JoeZiminski: yes agree, maybe we can have a guideline that says before making PR either move TODO to github issues or solve them. I did not make Github issue as its a very small thing, but was just been lazy!

JoeZiminski: actually, this one was complete but I do have a bad having of leaving TODO that feel too small to be an issue but too lazy to do in the moment.

datashuttle/utils_mod/rclone_utils.py
adamltyson: > unfortunately leads to some problems with compilation across systems ([discussed here](https://forum.rclone.org/t/platform-independent-use-of-librclone-python-api/33286) - the rlcone people are very helpful!).

Would using conda not solve these problems?

> However, this is a API that spins a HTTP server to use. This is nice as it means the job can be communicated to from other machines. However I don't know much about HTTP servers and was worried about it creating problems (vague, unspecified problems so would probably be okay) , so have avoided for now with an eye to extend in future. However, do you think it might be worth using this from the outset?

No idea to be honest. Happy to keep it simple for now.

datashuttle/configs.py
adamltyson: I think anything can be an issue. If it's tiny enough so that the title is sufficient context, then I think that's fine.

JoeZiminski:
Thanks for this, just to check my understanding on the package vendoring. In this case I would make a folder called bg-atlasapi in the project root, and within that folder copy the function (or module?) from bg-atlasapi? Agree sounds much nicer than current implemetnation

datashuttle/utils_mod/rclone_utils.py
JoeZiminski: I dont think so, as rather than the rclone executable the python wrapper calls librclone.so but rclone dont supply / distribute the compiled .so so it would require us to compile for all supported systems and ship them (I'm not quite sure why it does not call the rclone executable directly)

Issue #13, opened 2022-10-27
Add CLI


JoeZiminski:
Currently interface all through python API. Plan to make GUI but first make CLI

PR #20, opened 2022-11-01
Adding cli

JoeZiminski:
PR to add CLI (now using argparse).

The main chanes are [CLI module](https://github.com/neuroinformatics-unit/datashuttle/blob/adding_cli/datashuttle/command_line_interface.py) and [CLI tests](https://github.com/neuroinformatics-unit/datashuttle/blob/adding_cli/tests/test_command_line_interface.py), are mostly wrappers around the API. Docs take the API docstring, with some additonal help info on the CLI options (I was not sure about these e.g. (str) for datatype).

Currently one very sub-optimal thing is the handling of flags for bools (we will probably remove a lot of these options if we go more the currently preferred folder structure. So, it might be easiest to deal with this during the re-write for that):

For the configs, we currently have ```use_behav=True, use_ephys=True``` etc. These are on by default. For the CLI I see flags for bools are preferred, so these would need to be changed to ```dont_use_behav=False, dont_use_ephys=False```. I have done this for the only other such flag, ```make_ses_dirs```, turning it into ```dont_make_ses_dirs```, but it less readable. For these other config flags, I just kept them as optional arguments and process the strings directly e.g. ```--use_behav False``` with default argument True. This is not very optimal though.

#13


niksirbi:
An alternative for handling the default booleans that need to be flags in the CLI:
`dont_use_ephys = False` or similar expressions are indeed hard to parse for (human) minds because of the double negative involved. A human-friendlier alternative could be `skip_ephys` or `drop_ephys` or `omit_ephys`. Same boolean logic, but somehow more readable without the `not`.

tests/test_make_dirs.py
```diff
@@ -223,7 +223,7 @@ def test_make_sub_dir_no_tree(self, project):
         """
         Make sub directory only, check it is made and no lower level dirs exist
         """
-        project.make_sub_dir("ephys", "001", make_ses_tree=False)
+        project.make_sub_dir("ephys", "001", dont_make_ses_tree=True)
```
niksirbi, line 226: I suggest renaming to "skip_ses_tree"

tests/test_make_dirs.py
```diff
@@ -233,18 +233,7 @@ def test_make_sub_dir_with_ses_no_tree(self, project):
...
-        )
-        test_utils.check_and_cd_dir(ses_path)
-        assert glob.glob(join(ses_path, "*")) == []
-
-    def test_make_empty_ses_dir(self, project):
-        """
-        Make an empty sub directory, checking no lower level dirs exist
-        """
-        project.make_empty_ses_dir("ephys", "001", "001")
+        project.make_sub_dir("ephys", "001", "001", dont_make_ses_tree=True)
```
niksirbi, line 236: Same as above

datashuttle/datashuttle.py
```diff
@@ -125,7 +130,7 @@ def make_sub_dir(
         experiment_type: str,
         sub_names: Union[str, list],
         ses_names: Union[str, list] = None,
-        make_ses_tree: bool = True,
+        dont_make_ses_tree: bool = False,
```
niksirbi, line 133: suggest renaming to 'skip_ses_tree'

datashuttle/datashuttle.py
```diff
@@ -298,7 +308,7 @@ def write_public_key(self, filepath: str):
     def make_config_file(
```
niksirbi, line 308: The arguments that default to `None` should include `None` as a possible type, otherwise `mypy` complains with:
> error: Incompatible default for argument "{ARG_NAME}" (default has type "None", argument has type "str")

datashuttle/datashuttle.py
```diff
@@ -531,7 +563,7 @@ def _make_directory_trees(
         experiment_type: str,
         sub_names: Union[str, list],
         ses_names: Union[str, list],
-        make_ses_tree: bool = True,
+        dont_make_ses_tree: bool = False,
```
niksirbi, line 566: `skip_ses_tree`

datashuttle/command_line_interface.py
```diff
@@ -0,0 +1,662 @@
...
+    "\n\n"
+    "All command and argument names are matched to the API "
+    "documentation. "
+    "\n\n"
+    "----------------------------------------------------------------------"
+)
+
+parser = argparse.ArgumentParser(
+    prog="datashuttle",
+    usage="%(prog)s [PROJECT NAME]",
```
niksirbi, line 117: I am curious. Is there a specific reason to use the old printing style instead of f"{prog} [PROJECT NAME]"?

datashuttle/command_line_interface.py
```diff
@@ -0,0 +1,662 @@
...
+    "new_info", action="store", help="(str or bool) depending on option key"
+)
+
+
+# ------------------------------------------------------------------------------------------
+# Setup SSH
+# ------------------------------------------------------------------------------------------
+
+
+def setup_ssh_connection_to_remote_server(project, args):
```
niksirbi, line 309: why is args here if not accessed

datashuttle/datashuttle.py
JoeZiminski: That's strange mine does not complain, are you running by any change with ```--no-implicit-optional```? If I understand [this post](https://stackoverflow.com/questions/51055952/does-a-default-parameters-overwrite-type-hints-for-mypy) correctly it seems the default behaviour should be to implicitly add None as an optional type (unless --no-implicit-optional is passed). I vaguely remember looking into this argument before but I can't see it in the config files but may have forgotton it somewhere.

Good point and I think worth discussing as a team what our preffered approach is. Mine would be to keep as default to avoid clutter, but I see some people (e.g. the person in that thread) like things stricter so worth discussing.

niksirbi: Thar's strange. I simply ran `mypy -p datashuttle`. But perhaps our `mypy` installs are somehow configured differently somewhere.
I agree it's a good idea to discuss it as a team. Personally, I would prefer to avoid the clutter if possible

datashuttle/command_line_interface.py
JoeZiminski: Yes I wasn't sure about this, argparse passes the args with the linked .func attached, which could be any of the run functions involved in running CLI commands (e.g. check_name_processing). Couldnt figure out how to handle this nicely, but now used *args in these cases

datashuttle/command_line_interface.py
JoeZiminski: good question! it was just how I copied it from SO, but just tested it and new the new way does not work (as progs is not defined in this scope and I guess is handled internally from the string by argparse). Have added a comment!

tests/test_make_dirs.py
JoeZiminski: good shout, now this option has been removed from later versions

JoeZiminski:
Thanks Niko, I pushed some commits to address these points to #35. I think the only outstanding thing is the mypy option.

niksirbi:
Cool, I will quickly check if I find anything else and I will approve this for merging today,

datashuttle/datashuttle.py
niksirbi: I tested the `mypy` thing on a different machine and it finds no problems now

JoeZiminski: That's strange! I will keep an eye out for this problem in the automation tests. Thanks a lot for looking into this

datashuttle/command_line_interface.py
```diff
@@ -0,0 +1,662 @@
...
+# ------------------------------------------------------------------------------------------
+
+description = (
+    "----------------------------------------------------------------------\n"
+    "DataShuttle command line interface. "
+    "\n\n"
+    "datashuttle [PROJECT NAME] [COMMAND] [OPTIONS]"
+    "\n\n"
+    "To get detailed help for commands and their optional arguments, "
+    "\ntype python -m datashuttle [PROJECT NAME] [COMMAND] --help'"
```
niksirbi, line 101: On my local install `python -m datashuttle` doesn't work. However running `datashuttle` directly in the terminal works every time. Should the desctiption just say `datashuttle` everywhere instead of `python -m datashuttle`?

niksirbi:
When calling `datashuttle test-shuttle update_config --help`, I get:
`usage: datashuttle [PROJECT NAME] update_config [-h] option_key new_info`.
However, when trying to run it I figured out that the actual syntax it accepts is:
`usage: datashuttle [PROJECT NAME] update_config [-h] option_key=new_info`.

JoeZiminski:
Thats interesting, is this on linux? It is passing in test environment but maybe there is another case. In the terminal (windows) in pycharm I get the expected behaviour:

```
(datashuttle_conda) C:\fMRIData\git-repo\datashuttle>datashuttle test update_config local_path "hello"  
local_path has been updated to hello

(datashuttle_conda) C:\fMRIData\git-repo\datashuttle>datashuttle test update_config local_path="hello" 
usage: datashuttle [PROJECT NAME] update_config [-h] option_key new_info
datashuttle [PROJECT NAME] update_config: error: the following arguments are required: new_info
```

niksirbi:
Yes, it's on our linux workstations.
Specifically, if I run `datashuttle test-shuttle update_config sub_prefix use-ephys true`, I get:
```bash
usage: datashuttle [PROJECT NAME]
datashuttle: error: unrecognized arguments: true
```

niksirbi:
@JoeZiminski pleas ignore the above error, I was mistyping the command (as you can see from the above example). It works as intended

niksirbi:
Okay, I tested all CLI functionality (minus the ssh connections), and afaik it works as intended.
Sth to consider as potential enhancements from the user's perspective:
* It would be nifty to have a shorthand for `datashuttle test-shuttle upload_data --experiment_type all --sub_names all --ses_names all`, perhaps sth like `datashuttle test-shuttle upload_data  --all`. This would essentially be similar to the `git push` functionality. 
* It may make sense to have an alternative to `make_config_file`, where the user can supply a ready-made `config.yaml` they have made.

niksirbi (approved):

datashuttle/command_line_interface.py
JoeZiminski: thanks good spot!

JoeZiminski:
Thanks a lot Niko, I added some more doc to the update_config CLI because it is confusing what option format is been referred to. That's a nice idea about the --all flag, I will make an issue. Good point it is possible for the user to swap out the yaml file directly but they would need to drag it to the appropriate folder, which is hidden on some OS. Much better to supply a convenieince function where a path can be input. I will make issues for these!

PR #54, opened 2022-12-12
Move config convenience function

JoeZiminski:
PR to add a [convenience function](https://github.com/neuroinformatics-unit/datashuttle/blob/d454339ff5a8d84b47210e5bb7695047149e6179/datashuttle/datashuttle.py#L460) that allows user-supplied config file. This blew up slightly into a refactor of the config-checking. Now there is a [cannonical_configs](https://github.com/neuroinformatics-unit/datashuttle/blob/d454339ff5a8d84b47210e5bb7695047149e6179/datashuttle/utils_mod/canonical_configs.py#L1) utils class that holds specification on how the config must be formatted, and a checking function. Whenever a config is set for the first time, updated using update_config or passed as a user-supplied config file, this function is used to check the config is correct.

configs were also [refactored to only take a connection_method and remote_path argument](). Now if the user
wants to switch connection method, they will have to supply a new remote_path. In practice, I dont think people will often
need to switch the location of the remote. There may be a better way to handle this in future though. Also, it will be nice to refactor ssh options into an ssh_settings dict rather than separate arguments, but this could be done later.


datashuttle/utils_mod/canonical_configs.py
```diff
@@ -0,0 +1,188 @@
...
+    check_config_types(config_dict)
+
+    if list(config_dict.keys()) != list(canonical_dict.keys()):
+        utils.raise_error(
+            f"New config keys are in the wrong order. The"
+            f" order should be: {canonical_dict.keys()}"
+        )
+
+    if config_dict["connection_method"] not in ["ssh", "local_filesystem"]:
+        utils.raise_error("connection method must be ssh or local")
```
niksirbi, line 100: Discrepancy between the if statement and the error message: is it "local" or "local_filesystem"?

datashuttle/utils_mod/canonical_configs.py
```diff
@@ -0,0 +1,188 @@
...
+        utils.raise_error(
+            "remote_host_id and remote_host_username are "
+            "required if connection_method is ssh."
+        )
+
+
+def handle_cli_or_supplied_config_bools(dict_):
+    """
+    For supplied configs for CLI input args,
+    in some instances bools will as string type.
```
niksirbi, line 128: missing verb after will

datashuttle/utils_mod/canonical_configs.py
```diff
@@ -0,0 +1,188 @@
...
+
+def get_canonical_config_required_types():
+    """
+    The only permitted types for DataShuttle
+    config values.
+    """
+    required_types = {
+        "local_path": Union[str, Path],
+        "remote_path": Union[str, Path],
+        "connection_method": str,
```
niksirbi, line 49: because this is not any string but only takes one of two specific string values, I think you can use `typing.Literal`

niksirbi: But double-check the above, I'm not sure I've got the usage of `Literal` 100%

datashuttle/command_line_interface.py
```diff
@@ -326,6 +291,7 @@ def make_sub_dir(project, args):
     metavar="",
 )
 make_sub_dir_parser.add_argument(
+    "--experiment-type",
     "--experiment_type",
```
niksirbi, line 295: Why is this specified twice? Same for the two fucntion calls above this one.

datashuttle/configs.py
```diff
@@ -1,84 +1,95 @@
...
 
 
 class Configs(UserDict):
-    """ """
+    """
+    Class to hold the configs for DataShuttle operations.
+    The configs must match exactly the standard set
+    in utils.cannonical_configs. If updating these
+    configs, This should be done here. This is setup to be
+    make config settings explicit and provide easy checking
```
niksirbi, line 19: remove be

datashuttle/command_line_interface.py
```diff
@@ -98,12 +64,12 @@ def help(help_type):
     "datashuttle [PROJECT NAME] [COMMAND] [OPTIONS]"
     "\n\n"
     "To get detailed help for commands and their optional arguments, "
-    "\ntype python -m datashuttle [PROJECT NAME] [COMMAND] --help'"
+    "\ntype datashuttle [PROJECT NAME] [COMMAND] --help'"
     "\n\n"
     "On first use it is necessary to setup configurations. \ne.g."
-    "'python -m datashuttle [PROJECT NAME] make_config_file [OPTIONS]'"
+    "'datashuttle [PROJECT NAME] make_config_file [OPTIONS]'"
```
niksirbi, line 70: use dashes instead of underscores
```suggestion
    "'datashuttle [PROJECT NAME] make-config-file [OPTIONS]'"
```

datashuttle/datashuttle.py
```diff
@@ -295,34 +296,35 @@ def write_public_key(self, filepath: str):
     def make_config_file(
```
niksirbi, line 296: The NOTE in this function's docstring is outdated

datashuttle/command_line_interface.py
```diff
@@ -235,7 +192,7 @@ def update_config(project, args):
     option_key = kwargs["option_key"]
```
niksirbi, line 192: the option_key is currently required to be in the python conventional syntax, i.e. with underscor (e.g. remote_path).
Is there any way to also make it accept the alias (e.g. remote-path) when it is being callsed from the command line (given that in almost all other parts of the CLI both options work)?

datashuttle/datashuttle.py
```diff
@@ -458,6 +458,47 @@ def show_configs(self):
...
+        self, path_to_config: Union[Path, str], warn: bool = True
+    ):
+
+        path_to_config = Path(path_to_config)
+
+        utils.raise_error_not_exists_or_not_yaml(path_to_config)
+
+        if warn:
+            input_ = utils.get_user_input(
+                "This will overwrite the existing datashuttle config file."
```
niksirbi, line 471: This warning also appears when one tries to create a new project with a self-supplied config file.
As far as I understand, you would only want the warning to appear if config file is supplied for an existing project.

niksirbi (commented):
Great work @joeziminski. The introduced changes work well.

I left several comments, let me know if you need clarifications on any of them.

niksirbi (approved):

datashuttle/utils_mod/canonical_configs.py
JoeZiminski: That's awesome!

datashuttle/command_line_interface.py
JoeZiminski: Thanks for this, just realised the parser do not accept aliases :(

```
    make_config_file_parser.add_argument(
        "--show-transfer-progress",
        aliases=["--show_transfer_progress"],
        required=False,
        action="store_true",
        help=help("flag_default_false"),
    )
```
```
(datashuttle_conda) C:\Users\Joe\work\git-repos\datashuttle>datashuttle test update_config remote-path hello world
Traceback (most recent call last):
  File "C:\Users\Joe\miniconda3\envs\datashuttle_conda\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\Joe\miniconda3\envs\datashuttle_conda\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\Joe\miniconda3\envs\datashuttle_conda\Scripts\datashuttle.exe\__main__.py", line 4, in <module>
  File "C:\Users\Joe\work\git-repos\datashuttle\datashuttle\command_line_interface.py", line 836, in <module>
    parser = construct_parser()
  File "C:\Users\Joe\work\git-repos\datashuttle\datashuttle\command_line_interface.py", line 420, in construct_parser
    make_config_file_parser.add_argument(
  File "C:\Users\Joe\miniconda3\envs\datashuttle_conda\lib\argparse.py", line 1423, in add_argument
    action = action_class(**kwargs)
TypeError: __init__() got an unexpected keyword argument 'aliases'
```

datashuttle/command_line_interface.py
JoeZiminski: nice idea thanks!

PR #59, opened 2022-12-15
Change config path add convenience functions

JoeZiminski:
This PR adds a few small bits of functionality:

[Upload / download all convenience function (wrap upload_data() with "all" for all arguments](https://github.com/neuroinformatics-unit/datashuttle/blob/0e418f81429fe1345d4519ce0930e362c281c26f/datashuttle/datashuttle.py#L193)

Move the config path from appdirs (on windows) to the user home directory (now there for all platforms)

[Add a check that the user is not drying to create duplicates of subject or sessions](https://github.com/neuroinformatics-unit/datashuttle/blob/0e418f81429fe1345d4519ce0930e362c281c26f/datashuttle/utils/directories.py#L39)

Some refactoring, moving directory search functions from datashuttle.py to directories.py

### DIFF
--- a/datashuttle/datashuttle.py
+++ b/datashuttle/datashuttle.py
@@ -125,11 +125,65 @@ class DataShuttle:
         else:
             ses_names = self._format_names(ses_names, "ses")
 
+        self._check_no_duplicate_sub_ses_key_values(sub_names, ses_names)
+
         self._make_directory_trees(
             sub_names,
             ses_names,
             data_type,
         )
+
+    def _check_no_duplicate_sub_ses_key_values(
+        self, new_sub_names: List[str], new_ses_names: List[str]
+    ) -> None:
+        """ """
+        base_dir = self._get_base_and_top_level_dir("local")
+
+        if new_ses_names is None:
+            existing_sub_names = self._search_sub_or_ses_level(
+                base_dir, "local"
+            )
+            existing_sub_values = self.get_first_sub_ses_keys(
+                existing_sub_names
+            )
+
+            for new_sub in self.get_first_sub_ses_keys(new_sub_names):
+                if new_sub in existing_sub_values:
+                    utils.raise_error(
+                        f"Cannot make directories. "
+                        f"The key sub-{new_sub} already exists in the project"
+                    )
+
+        # for each subject, check session level
+        for sub in new_sub_names:
+            existing_ses_names = self._search_sub_or_ses_level(
+                base_dir, "local", sub
+            )
+            existing_ses_values = self.get_first_sub_ses_keys(
+                existing_ses_names
+            )
+
+            for new_ses in self.get_first_sub_ses_keys(new_ses_names):
+
+                if new_ses in existing_ses_values:
+                    utils.raise_error(
+                        f"Cannot make directories. "
+                        f"The key ses-{new_ses} for {sub} already exists in the project"
+                    )
+
+    def get_first_sub_ses_keys(self, all_names: List[str]) -> List[str]:
+        """
+        Assumes sub / ses name is in standard form with sub/ses label
+        in the second position e.g. sub-001_id-...
+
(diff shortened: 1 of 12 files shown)


datashuttle/datashuttle.py
```diff
@@ -183,6 +190,22 @@ def download_data(
...
+        """
+        self.upload_data("all", "all", "all")
+
+    def download_all(self):
+        """
+        Convenience function to download all data.
+        Alias for:
+            project.download_data("all", "all", "all")
+        """
+        self.download_data("all", "all", "all")
```
lauraporta, line 207: maybe we can have "all" "all" "all" to be more explicit in the future

datashuttle/utils/directories.py
```diff
@@ -30,31 +29,174 @@ def make_dirs(paths: Union[Path, List[Path]]) -> None:
...
+def search_data_dirs_sub_or_ses_level(
+    project, data_type_dirs, base_dir, local_or_remote, sub, ses=None
+):
+    """
+    Search  a subject or session directory specifically
+    for data_types. First searches for all folders / files
+    in the directory, and then returns any dirs that
+    match data_type name.
+
+    see _search_sub_or_ses_level() for inputs.
```
lauraporta, line 138: You might want to restyle the docstrings and make more explicit the parameters names and their type, and the return type

datashuttle/utils/ssh.py
```diff
@@ -116,6 +117,7 @@ def verify_ssh_remote_host(remote_host_id: str, hostkeys: Path) -> bool:
     get the server key and present when connecting
     for manual validation.
     """
+    transport: paramiko.Transport
```
lauraporta, line 120: why do you need to specify the type above if then you're using a `with` statement?

lauraporta (approved):

datashuttle/datashuttle.py
JoeZiminski: do you mean to change as below in the docstring?
```
        Alias for:
            project.download_data(sub_names="all",
                                               ses_names="all", 
                                               data_type="all")
```

datashuttle/utils/directories.py
JoeZiminski: Thanks this has type hints in a [newer version](https://github.com/neuroinformatics-unit/datashuttle/blob/6e03f8be840459480aeabecfd849648cbc4f4065/datashuttle/utils/directories.py#L278) and I've made the docstring clearer

datashuttle/utils/ssh.py
JoeZiminski: Good question I kept getting a strange mypy error 
```
datashuttle\utils\ssh.py:138: error: Self? has no attribute "connect"  [attr-defined]
datashuttle\utils\ssh.py:139: error: Self? has no attribute "get_remote_server_key"  [attr-defined]
Found 2 errors in 1 file (checked 1 source file)
```
which I couldnt seem to solve or find much support online, I wonder if it is a bug related to paramiko's type hints. explicitly defining the type here solved the issue (though not sure why)

datashuttle/datashuttle.py
lauraporta: It confuses me to pass the concept of the action of "downloading all" with a string. What else can you pass?
I see that then you use it in another method with an if statement `if ses_names not in ["all", ["all"]]:`. 
I suggest to refactor in the future with `download_all_subnames=True` etc.. in case there is the chance of this method to be part of the user facing API

JoeZiminski: I see this is a good point, currently you can pass different list / string e.g. `"sub-001"`, `["sub-001",, "sub_002", "all_non_ses"]` for various combinations of files, for the user this is not super-clear. In future hopefully 99% of use will be through TUI but I'll add a note on the relevant issue thanks

PR #64, opened 2022-12-20
Add logging

JoeZiminski:


datashuttle/utils/ds_logger.py
```diff
@@ -0,0 +1,114 @@
...
+def start(
+    path_to_log: Path, name: str, variables: Optional[List[Any]]
+) -> None:
+    """"""
+    fancylog.start_logging(
+        path_to_log,
+        package,
+        filename=name,
+        variables=variables,
+        verbose=False,
```
adamltyson, line 26: Would it be good to not hard code this to `False`? It may be useful to increase the amount of logging to the console when debugging.

datashuttle/utils/ds_logger.py
```diff
@@ -0,0 +1,114 @@
...
+        timestamp=True,
+        file_log_level="INFO",
```
adamltyson, line 28: Is there a reason for setting this as "INFO"? I think this means that "DEBUG" information (which should be the majority) wouldn't be saved to the log.

datashuttle/utils/ds_logger.py
```diff
@@ -0,0 +1,114 @@
...
+        write_git=False,
```
adamltyson, line 29: Sorry to raise so many fancylog questions, but I know the package well... Any reason to not write this if it's available? If someone is using a version that's not released, it's useful to get the relevant git commit straight from the log. The info should be in the setuptools_scm version, but I think this is easier.

datashuttle/utils/ds_logger.py
```diff
@@ -0,0 +1,114 @@
...
+        log_to_console=False,
```
adamltyson, line 30: Do you not want any info sent to the console, or are you printing as well as logging?

datashuttle/utils/ds_logger.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+    logging.info(f"Starting {name}")
```
adamltyson, line 32: I'm not sure this info message would be very clear to the user.

datashuttle/utils/utils.py
```diff
@@ -9,11 +13,30 @@
 # --------------------------------------------------------------------------------------------------------------------
 
 
-def message_user(message: Union[str, list]) -> None:
+def log(message: str) -> None:
+    logger = logging.getLogger("fancylog_")
+    logger.info(message)
```
adamltyson, line 18: I think I'm misunderstanding something, but why is everything logged as `INFO`? Typically I log at one of the [existing levels](https://docs.python.org/3/library/logging.html#logging-levels) (creating new ones if really necessary), then you can control what info goes where based on the level of severity. I.e. users typically only need to see high-level details in the console (`INFO`), but the developer probably wants as many details as possible in the log (`DEBUG`).

datashuttle/utils/utils.py
```diff
@@ -9,11 +13,30 @@
 # --------------------------------------------------------------------------------------------------------------------
 
 
-def message_user(message: Union[str, list]) -> None:
+def log(message: str) -> None:
+    logger = logging.getLogger("fancylog_")
```
adamltyson, line 17: You probably don't want the logger named `fancylog`. What if you have >1 package installed that are using fancylog?

datashuttle/utils/utils.py
```diff
@@ -9,11 +13,30 @@
...
+    logger.info(message)
+
+
+def log_and_message(message: str, use_rich: bool = False) -> None:
+    log(message)
+    message_user(message, use_rich)
```
adamltyson, line 23: `message_user` to me isn't super clear what it means. Something like `print_message_to_user` would be clearer.

datashuttle/utils/utils.py
```diff
@@ -31,7 +54,7 @@ def raise_error(message: str) -> None:
     raise BaseException(message)
 
 
-def get_appdir_path(project_name: str) -> Path:
+def get_appdir_path(project_name: str) -> Tuple[Path, Path]:
     """
     It is not possible to write to program files in windows
```
adamltyson, line 59: I thought all logs were saved in the project folder? I think we want this (or in `~/.datashuttle`), not in Program Files.

datashuttle/utils/rclone.py
```diff
@@ -42,28 +41,33 @@ def transfer_data(
...
         )
 
+    return output
+
 
 def setup_remote_as_rclone_target(
     cfg: Configs,
     rclone_config_name: str,
     ssh_key_path: Path,
+    log: bool = False,
```
adamltyson, line 70: Just a general question, why would you want the default to ever be to not log?

datashuttle/utils/ssh.py
```diff
@@ -110,34 +110,45 @@ def add_public_key_to_remote_authorized_keys(
...
+        client.get_host_keys().save(hostkeys_path.as_posix())
         success = True
     else:
         utils.message_user("Host not accepted. No connection made.")
         success = False
 
+    if log:
+        if success:
+            utils.log(f"\n{message}")
+            utils.log(f"\nHostkeys saved at:{hostkeys_path.as_posix()}")
```
adamltyson, line 148: Doesn't the logging module add a new line for you?

tests/tests_integration/test_logging.py
```diff
@@ -0,0 +1,404 @@
...
+        )
+        assert "Update successful. New config file:" in log
+
+    def test_logs_update_config(self, setup_project):
+
+        setup_project.update_config("remote_path", "test_path")
+
+        log = self.read_log_file(setup_project._logging_path)
+
+        assert "Starting update_config" in log
```
adamltyson, line 96: Is it worth having a "validation" log file in with the tests? May be easier to load & compare rather than checking line by line here.

tests/tests_integration/test_logging.py
```diff
@@ -0,0 +1,404 @@
...
+            "formatted_sub_names: ['sub-1_1', 'sub-002', 'sub-003', 'sub-004']"
+            in log
+        )
+        assert "formatted_ses_names: ['ses-123', 'ses-hello_world']" in log
+        assert "Made directory at path:" in log
+        assert (
+            os.path.join("test_logging", "local", "rawdata", "sub-1_1") in log
+        )
+        assert (
+            os.path.join(
```
adamltyson, line 142: I know it would need to be converted to a string at the end, but I would use Pathlib here for consistency. Mostly because familiarity with os.path will probably decline over time.

datashuttle/utils/utils.py
```diff
@@ -9,11 +13,30 @@
...
+    message_user(message, use_rich)
+
+
+def log_and_raise_error(message: str) -> None:
+    logging.error(f"\n\n{' '.join(traceback.format_stack(limit=5))}")
+    logging.error(message)
+    raise_error(message)
+
+
+def message_user(message: Union[str, list], use_rich=False) -> None:
```
adamltyson, line 32: I don't fully understand why this is necessary, rather than using logging.

adamltyson (approved):
Hey @JoeZiminski looks good. I've raised a few questions, but I've approved so you can get on with things while I'm away.

I'm afraid it's a line-by-line review rather than a comment on the general structure. I found it a bit hard to review having not been involved in many of the other PRs.

Some general things I wasn't clear on:

* Why are messages sent to the log, but then also printed to the console rather than logging to the console as well as the log file? This seems to add mental overhead rather than just using the logging module directly in the code.  I think this approach is needed to ensure errors are logged, but everything else should be handled by the logging module (and multiple handlers).

* Where exactly is the log saved during "normal" use, and what is it called? It needs to be super easy for the user to find the file and send to us if something goes wrong.

* What `variables` are saved to the log? From experience, I would make sure absolutely everything is logged (other than SSH keys etc). The more you can save the state of the software into the log file, the easier it is to debug later on.

pyproject.toml
```diff
@@ -16,6 +16,8 @@ dependencies = [
     "PyYAML",
     "requests",
     "rich",
+    # "fancylog",
+    "fancylog @ git+https://git@github.com/JoeZiminski/fancylog",
```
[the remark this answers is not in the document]
JoeZiminski: This is slightly outdated, was related [to this PR](https://github.com/adamltyson/fancylog/pull/18) and has been updated now thanks

datashuttle/utils/ds_logger.py
JoeZiminski: Sure so just to hard-code it to True? (or alternatively provide as option?)

datashuttle/utils/ds_logger.py
JoeZiminski: I think the only reason was that I did not understand the logging levels sufficiently, will edit this!

datashuttle/utils/ds_logger.py
JoeZiminski: Not at all very useful. I think I fundamentally misunderstood the purpose of the log. Previously I thought it would be mainly used to researchers to keep track of any operations they might have done and look through the history. Us having spoken further it is clear this is not the way logs are typically used and instead should contain as much information for debugging as possible. A few options were turned off to keep the log cleaner for the user, but this is the wrong approach. Will set `True`!

datashuttle/utils/utils.py
JoeZiminski: thanks response as above will fix

datashuttle/utils/rclone.py
JoeZiminski: I don't think you would, this is inconsistent also across the codebase (sometimes log=True default), will fix

tests/tests_integration/test_logging.py
JoeZiminski: This is nice idea, unfortunately there seems to be some cross-platform differences in the log output and so selected randomly some lines that are the same cross platform. It's true though that these are completely arbitrary lines. The bar here was 'is printing a log that is describing the correct process' but these tests could be stricter.

datashuttle/utils/utils.py
JoeZiminski: (discussed above)

datashuttle/utils/ds_logger.py
adamltyson: To use this option, I usually have some kind of "debug" mode in my code, that would set this to True, and also change some other settings (e.g. for data analysis s/w, saving intermediate files).

datashuttle/utils/ds_logger.py
adamltyson: I don't know if this is a general principle, but I use `logging.info()` sparingly to keep the console logs clean for the user, but then I send absolutely everything to the log file (with `logging.debug()` but also fancylog options) to aid (usually remote) debugging later.

datashuttle/utils/ds_logger.py
JoeZiminski: That's a nice idea, for now I have exposed it as a top-level argument on `datashuttle._start_log()` and set default for True for now, as I guess we will be doing a lot of debugging. In future this can now be more easily extended to incorporate a debug mode.

datashuttle/utils/ds_logger.py
JoeZiminski: thanks have changed to `logging.info(f"Starting logging for command {name}")`

datashuttle/utils/ds_logger.py
JoeZiminski: In the end I think I will continue with the current implementation to split the logging into the file and handle the print statements separately. The reason for this is to have a bit more flexibility in future when we might want to display messages to users in a different way (e.g. a pop-up box in a TUI).

adamltyson:
> In the end I continued to split the logging through fancylog while manually controlling the output to retain a bit more flexibility. Currently showing messages to use just uses print() but in future we might need to show them as TUI output. Also, maybe I am missing something but this also allows to cut down on a bit of verbose-ness when sending user messages.

Sounds sensible. I think it should be possible to adjust the logging to console, but as you say, there's likely to be multiple places we want to send this info. 

>Currently the logs are saved to the .datashuttle folder users local project folder tree (i.e. .datashuttle is in the same folder as rawdata. However actually, this might cause problems if they don't know about hidden folders? Maybe it is better to save in a dedicated logs folder in the local project top level dir?

Yeah I think you're right. The users may also want to look at them to be sure something has been synced. 

>No variables were saved but as you say this is very useful (and a nice feature in fancylog to take advantage of). I've [added this here](https://github.com/neuroinformatics-unit/datashuttle/pull/90/commits/696bd7a034246d214d5611de61b74860cccf30a7) to save the locals() of whatever command was passed to get the user arguments, and the datashuttle configs at the time.

Sounds good.

PR #68, opened 2023-01-04
Add docs

JoeZiminski:
Update documentation and add to website, inlcuding documentation, API and CLI auto-generated docs

This required some refactoring of command_line_interface.py as the docs builder expects the parser to be returned from a function. Some more function docstrings were added / edited as these are used in auto-build docs.

Otherwise, the majority of changes are in the new /docs folder.


docs/source/pages/get_started.md
```diff
@@ -0,0 +1,183 @@
...
+To get started, import DataShuttle and initialise the class with the project name. If using the command
+line this step is not necessary.
+
+```
+from datashuttle.datashuttle import DataShuttle
+
+project = DataShuttle("my_project")
+```
+
+DataShuttle helps to manage and transfer a project with many "local" machines all connected to a central "remote" machine.
```
adamltyson, line 33: The concept of a "project" isn't explained. I would explain it a bit, and link to the swc-blueprint docs for the user to read first.

docs/source/pages/get_started.md
```diff
@@ -0,0 +1,183 @@
...
+DataShuttle requires a path to the local project directory (this is typically empty on first use and should include the name of the project
+e.g. local_path="/path/to/my_project"). Also, remote project directory and method to connect to the remote project (either "local_filesystem" or
+"ssh").
+
+It is also required to specify the data types (e.g. behav, ephys, funcimg, histology) used on the local PC. If data_type flags are not set to True, it will not be possible
```
adamltyson, line 38: As above, the user may not understand what exactly the data types are, so they should look through swc-blueprint first.

docs/source/pages/get_started.md
```diff
@@ -0,0 +1,183 @@
...
+to create directories of these data types.
+
+If connection_method used is "ssh", it is necessary to also input the remote_host_id and remote_host_username configs.
```
adamltyson, line 41: I don't think this line would be easy to understand for most.

docs/source/pages/get_started.md
```diff
@@ -0,0 +1,183 @@
...
+use_ephys=True,
+use_behav=True,
+use_histology=True
+)
+```
+or equivalently using the command-line interface
+
+
+```
+datashuttle \
```
adamltyson, line 61: Can this be split over fewer lines? It's not obvious that it's a single CLI command.

docs/source/pages/get_started.md
```diff
@@ -0,0 +1,183 @@
...
+--use-ephys --use-behav --use-histology
+```
+
+Individual settings can be updated using update_config(), and an existing config file can be used instead using supply_config()
+
+### Setting up an SSH Connection
+
+Once configurations are set, if the "connection_method" is "ssh", the function setup_ssh_connection_to_remote_server() must be run to setup
+the ssh connection to the remote server. This will allow visual confirmation of the server key, and setup a SSH key pair. This means
+your password will have to be enterred only once, when setting up this connection.
```
adamltyson, line 78:
```suggestion
your password will have to be entered only once, when setting up this connection.
```

docs/source/pages/get_started.md
```diff
@@ -0,0 +1,183 @@
...
+
+
+## Making Directories
+
+Subject and session project directories can be made using the function make_sub_dir(). This function accepts a subject name (or list
```
adamltyson, line 83:
```suggestion
Subject and session project directories can be made using the function `make_sub_dir()`. This function accepts a subject name (or list
```

docs/source/pages/get_started.md
```diff
@@ -0,0 +1,183 @@
...
+of subject names), with optional session name and data type inputs. If no session or data type name is provided,
+an empty subject directory will be made at the top directory level.
+
+All subject or session names must be prefixed with "sub-" or "ses-" respectively, as according to SWC-BIDS. If these prefixes
+are not input, they will be automatically added. To make all data types (i.e. all datatypes for which use_<data_type> configuration is true),
+ data_type="all" can be set.
+
+Convenience tags (@TO@, @DATE@, @TIME@, @DATETIME@) for creating ranges of subjects, or automatically including
```
adamltyson, line 91: > (@TO@, @DATE@, @TIME@, @DATETIME@) 

I would explain this more

docs/source/pages/get_started.md
```diff
@@ -0,0 +1,183 @@
...
+```
+
+## Data Transfer
+
+Data transfer can be either from the local project to the remote project ("upload") or from the remote to local project("download"). Data
+transfers are primarily managed using the upload_data() and download_data() functions.
+
+By default, uploading or downloading data will never overwrite files when transferring data. If an
+existing file with the same name is found in the target directory, even if it is older, it will not be overwritten.
+All transfer activity is printed to the console and logged (see "Logging" below), which can be used to
```
adamltyson, line 145: > (see "Logging" below)

Can this be a link to the section below?

docs/source/pages/get_started.md
```diff
@@ -0,0 +1,183 @@
...
+will only transfer behavioral data type folders, for sessions 1 and 5 from all subjects.
+
+Along with the @TO@ flag, a wildcard @*@ flag can also be used in subject or session names.
+
+
+## Logging
+
+Detailed logs of all configuration changes, full paths to created directories and data transfers are logged
+to a .datashuttle directory in the local project directory. These logs are named
+with the command (e.g. make_config_file), date and time of creation.
```
adamltyson, line 178: Not the place to review this? But is there one log file per command? Does this not get unwieldy quickly? Why not just one rotating log file, or one per session/day?

docs/source/pages/get_started.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Convenience Functions
+
+Convenience functions can be used to quickly get relevant project information. See the API or CLI documentation
```
adamltyson, line 182: >  See the API or CLI documentation

Can these be linked?

adamltyson (changes requested):

adamltyson:
I left a load of comments on the `Get Started` section, but in general, I would refactor it a bit. It seems like the `Get Started` section is just a subset of `Full Documentation". I don't think "Get Started" has enough information to get them started. I would do something like:

* Introduction - all the basic info of why DataShuttle exists, what the main concepts are, and links to SWC-BluePrint so that the user can understand about projects etc
* Get Started - A basic walkthrough of either API/CLI to start transferring data
* Full Docs - All the information needed to use every feature
* API & CLI docs - as they are currently, a reference of specific functions/commands

PR #69, opened 2023-01-17
Testing ssh and transfer

JoeZiminski:
This PR contains a lot of refactoring, the review can be targeted to some new functionality:

1) The logic for transferring data was changed so that one call to rclone is made, with a list of files to transfer, rather than making one call. This speeds up data transfer, if we ever want to test against a one-call-per-folder implementation, we can pull the old version from the git history. This new implementation is at [data transfer.py](https://github.com/neuroinformatics-unit/datashuttle/blob/testing_ssh_and_transfer/datashuttle/utils/data_transfer.py)

2) [Tests](https://github.com/neuroinformatics-unit/datashuttle/blob/testing_ssh_and_transfer/tests/tests_integration/test_ssh_setup.py) for the initial setup of SSH connectivity. Also, setup for tests is described in [conftest.py](https://github.com/neuroinformatics-unit/datashuttle/blob/testing_ssh_and_transfer/tests/conftest.py)

3) Testing SSH file transfer. The new tests work by specifying all folders / files to generate in a [big table](https://github.com/neuroinformatics-unit/datashuttle/blob/testing_ssh_and_transfer/tests/tests_integration/test_file_conflicts_pathtable.py). The [main tests](https://github.com/neuroinformatics-unit/datashuttle/blob/testing_ssh_and_transfer/tests/tests_integration/test_ssh_file_transfer.py) then take this table, create the project and then transfer combinations of different sub / ses / datatype. These can also be tested over SSH, but require setting up the SSH connection (this will always be off, unless we set it on our testing PC).

4) Some more rclone settings were exposed to the user, and these can be set in the config file. Tests for [these are here](https://github.com/neuroinformatics-unit/datashuttle/blob/d06cfc6293fa91c347829c853911481d7d4aa287/tests/tests_integration/test_filesystem_transfer.py#L352)

Notes:

New keys for transferring data that is not within a data_type folder (as discussed in #70), - "all_non_sub", "all_non_ses", ... (full list [here](https://github.com/neuroinformatics-unit/datashuttle/blob/d06cfc6293fa91c347829c853911481d7d4aa287/datashuttle/configs/canonical_directories.py#L75). These keys handle the case where users might want to transfer files that do not lie in sub / ses / datatype folders. Otherwise, "all" will push every file and folder from rawdata/ down. They are explained in some new docs (`documentation.md` in the diff)


tests/conftest.py
```diff
@@ -6,7 +6,8 @@
 
 def pytest_configure(config):
     pytest.ssh_config = SimpleNamespace(
-        TEST_SSH=False,
+        TEST_SSH=True,
+        PASSWORD_FILE=r"C:\Users\Joe\temp_pass.txt",  # don't store this on github!
```
adamltyson, line 10: Should authentication not be carried out with ssh keys not passwords?

JoeZiminski: Yes actually, for nearly all tests this is okay. There is a test for setting up the SSH key-pair, which is why I added this. I thoguht I could get away with it as a temp workaround, but having already accidently pushed a password to github and had to change a load of passwords because of it, I now think its a terrible idea 🥲 , at least to have the file in the repo. Maybe it is safest just to scrap the test for SSH key pair setup, and rely on the keys for all over tests.

tests/tests_integration/test_file_conflicts_pathtable.py
```diff
@@ -0,0 +1,38 @@
+# flake8: noqa
+# fmt: off
+
+from pathlib import Path
+
+import pandas as pd
+
+
+def get_pathtable(base_dir):
```
adamltyson, line 9: Would it be easier to store this in a csv or something? It's hard to parse as code.

adamltyson (approved):

PR #100, opened 2023-05-09
[Testing] Setup SSH tests to cluster (internal testing only)

JoeZiminski:
This PR mixes up two new required types of testing 
1) using a docker image to test SSH - anyone with docker / singularity should be able to run these tests
2) a very specific test to ceph, this code shouldn't be in datashuttle, maybe we can fork it or find a better way.

The image-route will be implemented in another PR. This PR will be stripped of image-related things and when we extend our internal testing can be used as a base. But it is unlikely this PR will be merged into main.


adamltyson:
@JoeZiminski is this PR still useful? It's nearly two years old!

JoeZiminski:
2 years 😱 sadly it is still on the TODO alongisde #208, but I'm hoping to include in next release 🙏

JoeZiminski:
On further consideration, I don't think there is much point in these tests if we are already testing the ssh connection on linux using docker. Will close!

PR #106, opened 2023-05-15
Add short names for cli optional args

JoeZiminski:
Add single-char options e.g. "-i" for optional CLI arguments. 

However, in some cases I broke the single-character rule discussed [here](https://stackoverflow.com/questions/18270133/command-line-options-should-short-options-be-restricted-to-1-character) in particuarly, for `--histology`, `-h` conflicts with help I think so went for `-hi`. Also I used `-su` and `-se` for `--sub-names` and `--ses-names` respectively.

However, maybe single-char alternatives can be used if single-char short arguments is an important convention.

### DIFF
--- a/datashuttle/command_line_interface.py
+++ b/datashuttle/command_line_interface.py
@@ -418,6 +418,7 @@ def construct_parser():
     make_config_file_parser.add_argument(
         "--remote-host-id",
         "--remote_host_id",
+        "-i",
         required=False,
         type=str,
         help="(str)",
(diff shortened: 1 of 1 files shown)


adamltyson:
Might be worth asking @neuroinformatics-unit/neuroinformatics-team. IMO single-char alternatives would be good for anything users may type a lot. I think it's fine to come up with alternatives because:
* Lots of packages do it, there's only 26 (english) characters
* There's nothing stopping us having alternatives that are easier to remember.

niksirbi:
I'd agree with Adam here. We don't need single-character aliases for every argument, just for the often used ones. Also, we don't have to always use the first letter of the word (many programs violate this), as long as what we use makes some sense.
That said, I would not break universal conventions (e.g. I would keep `-h` for help, `-v` for version, etc.)

lauraporta (approved):

PR #115, opened 2023-05-23
Introduce setters, handle make_sub_folder and data transfer for additional top level folders (e.g. derivatives, code, analysis)

JoeZiminski:
This PR introduces handling of top-level folders in addition to `rawdata`, i.e. `derivatives`, `code` and `analysis`. In retrospect  it has gone slightly out of scope and introduces two related but distinct functionalities:

1) Set and show the current working top-level folder (e.g. `rawdata`, `derivatives`)
2) Implement functionality to transfer the entire project (i.e. the contents of all top-level-folders).

With #135, the current working top level folder is shown on project initialisation (API), when setting the top-level folder, and a subset of CLI commands.

Details
-------

The working `top_level_folder` can now be set using `set_top_level_folder` and shown with `show_top_level_folder`. The working top level folder is saved to the `persistent_settings`, so when the next session is started, the working top level folder is maintained.

All previous methods `make_sub_folders`, `upload_data` / `download_data`, `upload_all` / `download_all` still work only on the working top level folder. For example if you are in `derivatives` and choose `upload_all`, only the contents of derivatives will be uploaded.

As such, this PR introduces two new methods, `upload_entire_project()` and `download_entire_project()` and associated CLI. This will upload or download everything in every top level folder.

I am not sure if this naming scheme is confusing. Maybe `upload_all` should become `upload_working_top_level_folder` or something, but it is hard to think of a name that is concise but informative.

@neuroinformatics-unit/neuroinformatics-team it would be good to get some general feedback on this PR as it covers a topic we have all discussed together, and is quite critical for how datashuttle will be used.

### DIFF
--- a/datashuttle/command_line_interface.py
+++ b/datashuttle/command_line_interface.py
@@ -934,6 +934,7 @@ def construct_parser():
         description=process_docstring(
             DataShuttle.show_top_level_folder.__doc__
         ),
+        help="",
     )
     show_top_level_folder_parser.set_defaults(func=show_top_level_folder)
 
--- a/datashuttle/command_line_interface.py
+++ b/datashuttle/command_line_interface.py
@@ -225,6 +225,15 @@ def upload_all(*args: Any) -> None:
     project.upload_all()
 
 
+# Upload Entire Project -------------------------------------------------------
+
+
+def upload_entire_project(*args: Any) -> None:
+    """"""
+    project = args[0]
+    project.upload_entire_project()
+
+
 # Download Data ---------------------------------------------------------------
 
 
@@ -250,6 +259,15 @@ def download_all(*args: Any) -> None:
     project.download_all()
 
 
+# Download Entire Project -----------------------------------------------------
+
+
+def download_entire_project(*args: Any) -> None:
+    """"""
+    project = args[0]
+    project.download_entire_project()
+
+
 # Upload Project Folder or File -----------------------------------------------
 
 
@@ -265,7 +283,7 @@ def upload_project_folder_or_file(project: DataShuttle, args: Any) -> None:
     )
 
 
-# Download Project Folder or File ------------------------------------------------
+# Download Project Folder or File ---------------------------------------------
 
 
 def download_project_folder_or_file(project: DataShuttle, args: Any) -> None:
@@ -277,6 +295,20 @@ def download_project_folder_or_file(project: DataShuttle, args: Any) -> None:
         project.download_project_folder_or_file,
         kwargs.pop("filepath"),
         **kwargs,
+    )
+
+
+# Set Top Level Folder or File ------------------------------------------------
+
+
+def set_top_level_folder(project: DataShuttle, args: Any) -> None:
+    """"""
+    kwargs = make_kwargs(args)
+
+    run_command(
+        project,
+        project.set_top_level_folder,
+        kwargs["folder_name"],
     )
 
 
@@ -340,13 +372,22 @@ def show_logging_path(*args: Any) -> None:
     project.show_logging_path()
 
 
-# Show Local Tree ------------------------------------------------------------
+# Show Local Tree -------------------------------------------------------------
 
 
 def show_local_tree(*args: Any) -> None:
     """"""
     project = args[0]
     project.show_local_tree()
+
+
+# Show Top Level Folder -------------------------------------------------------
+
+
+def show_top_level_folder(*args: Any) -> None:
+    """"""
+    project = args[0]
+    project.show_top_level_folder()
 
 
 # Show Next Sub Number -------------------------------------------------------
@@ -384,7 +425,7 @@ def check_name_formatting(project: DataShuttle, args: Any) -> None:
 
 # Supply Own Config -----------------------------------------------------------
 
-#
+
 def supply_config_file(project: DataShuttle, args: Any) -> None:
 
     kwargs = make_kwargs(args)
@@ -648,6 +689,19 @@ def construct_parser():
     )
     upload_all_parser.set_defaults(func=upload_all)
 
+    # Upload All
+    # -------------------------------------------------------------------------
+
+    upload_entire_project_parser = subparsers.add_parser(
+        "upload-entire-project",
+        aliases=["upload_entire_project"],
+        description=process_docstring(
+            DataShuttle.upload_entire_project.__doc__
+        ),
+        help="",
+    )
+    upload_entire_project_parser.set_defaults(func=upload_entire_project)
+
     # Download Data
     # -------------------------------------------------------------------------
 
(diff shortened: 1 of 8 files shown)


niksirbi (commented):
I have two major comments on this one. These are just my purely subjective opinions, so definitely subject to modification.

## Top-level folders
I think the only top-level folders we should allow users to set should be `rawdata` and `derivatives`. These are the only two folders in which it makes sense to build sub-ses-datatype trees and to use the other functionalities of datashuttle. It's unlikely that the contents of `analysis` or `code` will follow the same structure. So from the user's POV, they only work in either `rawdata` mode or in `derivative` mode.

To allow users to transfer code/analysis and their contents, I propose modifying the `upload-project-folder-or-file` command, so that it takes any path relative to the project root. Thus, if I want to upload `code` and its contents, I can do so via `datashuttle my-project upload-project-folder-or-file code/**` In this way, the `upload-project-folder-or-file` and its download counterpart can be used to transfer any arbitrary project content regardless of the top-level folder setting. Perhaps I would rename this command to `upload-specific-folder-or-file` and would explicitly state that the path it takes as argument is relative to the project root.

## Confusing command names
I agree that the `upload_all` vs `upload_entire_project` is a confusing distinction to make, and renaming the latter to `upload_working_top_level_folder` is more explicit (however a bit cryptic in meaning).

perhaps the solution is to have only the commands 'upload_all' and 'download_all' which should take a positional argument of either [rawdata, derivatives, entire_project].
So the usage examples would be
```bash
datashuttle my-project upload-all rawdata
datashuttle my-project download-all derivatives
```

Adopting the two above points would mean that the following commands don't care about the top-level folder persistent setting.
```bash
upload-all
download-all
upload-project-folder-or-file
download-project-folder-or-file
```

JoeZiminski:
Hey @niksirbi thanks a lot for the review, I think this is a good point about `code` and `analysis`. It is probably a bit confusing to have the option to create a subject / session folder tree in the `code` repository. Its also a nice idea to be able to pass the top-level folder to transfer, we could actually open this up so that any top-level folder can be transferred regardless of name (as long as it exists). For example based on your suggestions we could have:

```
upload-entire-project  - uploads EVERYTHING in the project folder, irregardless of folder name (e.g. `rawdata`, `derivatives`, `code`, `some_random_folder`, `some_random_file`.
upload-top-level-folder rawdata - uploads the entire top-level folder specified (this doesn't have to be rawdata or derivatives). This is basically like `upload_all` but overrides the current directory and can be used on a top-level-folder with any arbitary name. 
upload-all - uploads the entire current working top level folder.
upload-data - upload subsets but relative to current working top level folder.
upload-specific-folder-or-file - uploads specific folder or file relative to working top level folder.
```
then only `rawdata` and `derivatives` can be set as the working top level folder. In the docs, things could be presented from a 'top-level working folder' perspective, and focus on `upload-all`, `upload-data`, `upload-specific-folder-or-file`. Then there could be a small note on `upload-top-level-folder-rawdata` or `upload-entire-project` in the case the working-folder approach needs to be overriden for custom folders (or convenience, if working in `rawdata` but want to upload all `derivatives`).

However, this looses some functionality compared to your suggestion, as it is not possible to upload specific files or folders in a non-`rawdata or non-`derivatives` folder. I like the idea of having the option to pass a top-level folder to these more flexible data transfer arguments, but am worried about mixing the two methods (i.e. passing an argument vs using the top level folder) in case it is confusing. Maybe something like a explicit `override_top_level_folder` could be added to `upload_all`, `upload_data`, `upload_specific_folder_or_file`? For now I think we could hold these back and then implement if there is demand for specific transfer of non-`rawdata` and non-`derivatives` files?

niksirbi:
```
upload-entire-project  - uploads EVERYTHING in the project folder, irregardless of folder name (e.g. `rawdata`, `derivatives`, `code`, `some_random_folder`, `some_random_file`.
upload-top-level-folder rawdata - uploads the entire top-level folder specified (this doesn't have to be rawdata or derivatives). This is basically like `upload_all` but overrides the current directory and can be used on a top-level-folder with any arbitary name. 
upload-all - uploads the entire current working top level folder.
upload-data - upload subsets but relative to current working top level folder.
upload-specific-folder-or-file - uploads specific folder or file relative to working top level folder.
```

I am fine with this scheme for now.
However, if we introduce the `upload-top-level-folder <FOLDER_NAME>` functionality, do we even need `upload-all`?
To my understanding, the following three commands will do (almost) the same thing (assuming we are in 'rawdata' mode):
```bash
upload-top-level-folder rawdata
upload-data --sub-names all --ses-names all --data-type all
upload-all
```
We had initially introduced `upload-all` as a convenience option (to avoid typing the 2nd command), but if `upload-top-level-folder` is there, it seems redundant. Removing it will also spare us the possible confusion with `upload-entire-project`.

Therefore my proposed scheme would be:
```bash
# Upload all project contents
upload-entire-project

# Upload the entire top-level directory with all its contents (`rawdata`, `derivatives`, or whatever else may be there)
upload-top-level-folder <FOLDER-NAME>

# The normal day-to-day upload method, that can work in 'rawdata' or in 'derivatives' mode, depending on the setting
upload-data --sub-names <SUB_NAMES> --ses-names <SES_NAMES> --data-type <DATA_TYPE>

# Upload specific path relative to either 'rawdata' or 'derivatives' depending on the setting (stays as is)
upload-specific-folder-or-file <PATH>
```

Sorry for dragging this on, but as you yourself mentioned, it's quite a consequential UI decision. Ultimately real user feedback will be valuable for this.
Let me know what you think.

JoeZiminski:
Thanks a lot Niko, not at all it's definitely important to get this right (or as right as possible subject to user feedback).

I agree actually there is too much redundancy in the scheme I proposed. I like your scheme, my only worry is on mixing explicitly passing the top-level folder and abstracting it to the set-top-level-folder scheme. Overall the three (incompatible) aims in general are:

1) allow the user to conveniently transfer all data without having to specify the top-level folder each time (e.g., if for 99% of the time they are in `rawdata`).
2) keep the API that determines how top level folders are transferred consistent (e.g. either set with top-level working directory or typing out during the transfer call). Considering 1 & 2, this means the top-level folder approach only is required.
3) provide a flexible API that allows transfer of arbitrary folder types. Considering 3 & 2, all transfer calls should take a top-level folder. But this breaks 1 :'D

As one of these three has to give, for the initial release, maybe we can drop `code` and `analysis` (and any arbitrary top level folder) support). We have `upload_data`, `upload_all` and `upload_entire_project` (that work as you described, except `upload_all` does not take the top-level-folder argument. Only `rawdata` and `derivatives` are supported, and are set as the top-level-working folder.

Undoubtedly people will request more flexibility eventually, but by this time how best to extend the functionality will hopefully be clearer based on feedback (e.g. maybe my worry for point 2, people will not actually find the mixed scheme confusing), and it also keeps things simpler for the initial release.

niksirbi:
Good way of pinning down the problem with the 3 incompatible (for now) aims.

> As one of these three has to give, for the initial release, maybe we can drop `code` and `analysis` (and any arbitrary top level folder) support). We have `upload_data`, `upload_all` and `upload_entire_project` (that work as you described, except `upload_all` does not take the top-level-folder argument. Only `rawdata` and `derivatives` are supported, and are set as the top-level-working folder.

I agree this is the best course of action for now. I'll approve this PR and let you tidy up things and merge when ready.

niksirbi (approved):

PR #120, opened 2023-05-30
Create persistent top level folder setting

JoeZiminski:
Create a `.yaml` file in the `.datashuttle` config folder that holds persistent settings. These are settings that will be persistent
across different sessions of the programme.

Initially they are used to hold and update the `top_level_folder` (e.g. `rawdata`, `derivatives`). This PR is required such that #115 can build on top of it.


adamltyson (approved):
LGTM

PR #128, opened 2023-05-31
Update docs deployment workflow

niksirbi:
Changed the Sphinx docs workflow so that deployment only happens when a version is tagged (same trigger as for PyPI release).


JoeZiminski:
Thanks @niksirbi! If it makes sense to, I will merge this PR after the upcoming PR that will fix up the docs. Then we can have a pretty good version up, with final version pushed with the tag (and then proper documentation versioning from there).

PR #135, opened 2023-05-31
Add print start up message as well as top-level-folder state to relevant methods.

JoeZiminski:
This PR sets a 'welcome' message on project initialisation. This at the moment only shows the current top-level working directory. However, we don't want to show this every time a CLI command is run (the CLI works by initialising the API and running the required function). So, this is turned off for CLI calls. 

Also, the `show_top_level_folder` method is called when all relevant functions are run (`make_sub_folders` and upload / download transfer methods that rely on the working top level folder).

### DIFF
--- a/datashuttle/command_line_interface.py
+++ b/datashuttle/command_line_interface.py
@@ -1025,6 +1025,50 @@ parser = construct_parser()
 # -----------------------------------------------------------------------------
 
 
+def should_show_startup_message(func_name: str) -> bool:
+    """ """
+    show_message_funcs = [
+        "make_config_file",
+        "update_config",
+        "supply_config_file",
+        "make_sub_folders",
+        "upload_data",
+        "upload_all",
+        "upload_entire_project",
+        "download_data",
+        "download_all",
+        "download_entire_project",
+        "upload_project_folder_or_file",
+        "download_project_folder_or_file",
+    ]
+
+    dont_show_message_funcs = [
+        "setup_ssh_connection_to_central_server",
+        "set_top_level_folder",
+        "show_local_path",
+        "show_datashuttle_path",
+        "show_config_path",
+        "show_central_path",
+        "show_configs",
+        "show_logging_path",
+        "show_local_tree",
+        "show_top_level_folder",
+        "show_next_sub_number",
+        "show_next_ses_number",
+        "check_name_formatting",
+    ]
+
+    assert func_name in show_message_funcs + dont_show_message_funcs, (
+        "func name not found. Make sure to add it to the list above"
+        "if implementing a new CLI command."
+    )
+
+    if func_name in show_message_funcs:
+        return True
+    else:
+        return False
+
+
 def main() -> None:
     """
     All arguments from the CLI are collected and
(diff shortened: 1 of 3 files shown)


datashuttle/command_line_interface.py
```diff
@@ -1046,6 +1046,7 @@ def main() -> None:
     These command functions (all defined above) will process
     the CLI arguments and then call the appropriate API function
     through run_command().
+
```
alessandrofelder, line 1049:
```suggestion
```
(Don't think there's a reason for this extra line?)

datashuttle/datashuttle.py
```diff
@@ -70,7 +70,7 @@ class DataShuttle:
                    see the path to this folder.
```
alessandrofelder, line 70: Worth documenting the second argument here too?

tests/tests_integration/test_ssh_file_transfer.py
```diff
@@ -18,7 +18,7 @@ class TestFileTransfer:
     @pytest.fixture(
         scope="class",
         params=[  # Set running SSH or local filesystem
-            #     False,
+            False,
```
alessandrofelder, line 21: I'm confused about why this change was made (and I can't for the life of me figure out where this is used? :rofl: ). 
Does `False` mean SSH or local filesystem or something completely different?
Maybe worth clarifying the comment or adding another comment?

alessandrofelder: (As a side note, and assuming I guess correctly that this change is unrelated to the other changes in the PR which all relate to displaying a message, maybe this should have gone in with a different PR?)

alessandrofelder (approved):
Largely looks good to me. Left a couple of minor points to consider before merging.
Do I understand correctly that we 
- never want to show the message when a `Datashuttle` object is constructed via the CLI
- always want to show the message when a `Datashuttle` object is constructed _not_ via the CLI
?

tests/tests_integration/test_ssh_file_transfer.py
JoeZiminski: Thanks Alessandro, this is very obscure and I've added some docs in the docstrings, let me know if this makes sense:

```
        - Pytest params - The `params` key sets the
        `params` attribute on the pytest `request` fixture.
        This attribute is used to set the `testing_ssh` variable
        to `True` or `False`. In the first run, this is set to
        `False`, meaning local filesystem tests are run. In the
        second run, this is set with a pytest parameter that is
        `True` (i.e. SSH tests are run) but is skipped if `TEST_SSH`
        in `ssh_config` (set in conftest.py` is `False`.
   ```
   
   You are correct this change should not really be in the PR, it is a consequence of me breaking the cardinal rule of 'don't comment out important stuff when testing because you'll forget to uncomment it' in a previous PR. A good reminder to myself never to do this! 😅.

JoeZiminski:
Thanks a lot @alessandrofelder, that is correct this is not displayed for the CLI. The reason is that it will be shown for every command which could be annoying (e.g. if using `show_local_path` its not necessary to see the entire welcome message. That being said, from the CLI there is no way to get this information. For now, this isn't a problem as it only prints the top level folder which can be accessed by `show-top-level-folder`. But it might be worth exposing a dedicated shower command if more things are added.

JoeZiminski:
Added a couple for fixed related to this PR in #161

PR #140, opened 2023-06-06
Rename transfer data functions.

JoeZiminski:
Further to #115, it will be nice to have the purpose of each transfer function (e.g. `upload_all`) very clear, which is not trivial. Currently (using examples involving `upload` only), `upload_data`, `upload_all` `upload_project_file_or_folder` are relative to the top-level working directory. `upload_entire_project` just uploads everything in `rawdata` and `derivatives`.

This PR is an attempt to make the naming scheme a little clearer. `upload_data` was shorted to `upload`. (This is nicer for API, but will make future searches a pain).

 `upload_all` was changed to `upload_working_folder`, as `all` is ambigious now that it is also possible to upload all top level folders. 

`upload_project_file_or_folder` was renamed to `upload_specific_file_or_folder`.

I'm not 100% sure on this and am happy to scrap this PR in case the namings are still not clear. What do you think @neuroinformatics-unit/neuroinformatics-team?

### DIFF
--- a/datashuttle/command_line_interface.py
+++ b/datashuttle/command_line_interface.py
@@ -203,7 +203,7 @@ def make_sub_folders(project: DataShuttle, args: Any) -> None:
 # Upload Data -----------------------------------------------------------------
 
 
-def upload_data(project: DataShuttle, args: Any) -> None:
+def upload(project: DataShuttle, args: Any) -> None:
     """"""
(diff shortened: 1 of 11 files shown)


lauraporta (approved):
It looks good to me, names are more self-explanatory now , substituting `all` was a good idea

PR #148, opened 2023-06-07
[Validation] Validate the length of subject or session values.

JoeZiminski:
This PR checks that the length of `sub` and `ses` values are consistent in the project. Initially, this was focused on checking the number of leading zeros, but this was not correct. The most recent version checks simply the length of the `sub-` or `ses-` key value, which is all we really care about. 

Currently this is tested within a new `validate_project` function. Ideally, this would be tested whenever a new folder is created, so these things are caught initially. However, this will require some integration with other PRs and so is left for completion with #249.

The PR also centralises the logic to grab all subject and session names from across the project (i.e. from this local machine and the central server).

In total it:
- introduces the `validate_project()` function for API and CLI. Currently this function just validates the number of leading zeros, but will do more in future (#249)
- Introduces `get_all_sub_and_ses_names()` and `get_local_and_central_sub_or_ses_names()` functions to get the subject and session names from local machine + the central machine. These functions will form the basis for most future validation routines.
- Adds three functions in `formatting.py` to check the sub and session value lengths across the project. 
- Adds tests for this functionality.

### DIFF
--- a/datashuttle/command_line_interface.py
+++ b/datashuttle/command_line_interface.py
@@ -363,6 +363,15 @@ def show_configs(*args: Any) -> None:
     project.show_configs()
 
 
+# Validate Project  ----------------------------------------------------------
+
+
+def validate_project(*args: Any) -> None:
+    """"""
+    project = args[0]
+    project.validate_project()
+
+
 # Show Logging Path ----------------------------------------------------------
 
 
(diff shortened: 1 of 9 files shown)


niksirbi (approved):
Looks good @JoeZiminski. I left a few suggestions, feel free to follow them or not.

There is also an empty `validation.py` file in utils, that's probably not intended?

JoeZiminski:
Thanks @niksirbi! This is much nicer approach, less error prone to return as dict

PR #171, opened 2023-06-19
Update tooling to use ruff et al.

adamltyson:
@JoeZiminski I've updated the tooling to match the cookiecutter template. I've tried to copy over the config from flake8 as best I could, but feel free to edit.


JoeZiminski (approved):
Thanks @adamltyson! Apologies completely forgot about this, Ruff is awesome.

PR #197, opened 2023-10-04
Add verbose option to searching folders functions.

JoeZiminski:
This PR introduces a `verbose` option to the folders that search for files and folders in the local or remote paths. By default, these will log if no path is found at the passed `search_string`. However in #148 , paths are searched in a blind manner and it is not known a priori if they will exist, so this message is very annoying. 

This PR is a blocker for #148 and should be merged before it.

### DIFF
--- a/datashuttle/utils/folders.py
+++ b/datashuttle/utils/folders.py
@@ -447,7 +447,8 @@ def search_for_folders(  # TODO: change name
     local_or_central : "local" or "central"
     search_path : full filepath to search in
     search_prefix : file / folder name to search (e.g. "sub-*")
-    verbose : If `True`, if a search folder cannot be found, a message
+    verbose : If `True`, when a search folder cannot be found, a message
+          will be printed with the missing path.
               will be printed with the un-found path.
     """
     if local_or_central == "central" and cfg["connection_method"] == "ssh":
--- a/datashuttle/utils/folders.py
+++ b/datashuttle/utils/folders.py
@@ -240,6 +240,7 @@ def search_sub_or_ses_level(
     sub: Optional[str] = None,
     ses: Optional[str] = None,
     search_str: str = "*",
+    verbose: bool = True,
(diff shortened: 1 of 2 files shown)


datashuttle/utils/folders.py
```diff
@@ -259,12 +260,15 @@ def search_sub_or_ses_level(
...
-    ses: either a session name (string) or None, This must not
+    ses : either a session name (string) or None, This must not
         be a session name if sub is None. If provided (with sub)
         then the session folder is searched
 
-    str: glob-format search string to search at the
+    str : glob-format search string to search at the
         folder level.
+
+    verbose : If `True`, if a search folder cannot be found, a message
```
niksirbi, line 270: I was confused by the wording, is this better?
```suggestion
    verbose : If `True`, when a search folder cannot be found, a message
          will be printed with the missing path.
```

datashuttle/utils/folders.py
```diff
@@ -437,16 +446,22 @@ def search_for_folders(  # TODO: change name
     local_or_central : "local" or "central"
     search_path : full filepath to search in
     search_prefix : file / folder name to search (e.g. "sub-*")
+    verbose : If `True`, if a search folder cannot be found, a message
```
niksirbi, line 449: Same here:
```suggestion
    verbose : If `True`, when a search folder cannot be found, a message
          will be printed with the missing path.
```

niksirbi: Same for other occurrences of this docstring as well (if you prefer my wording that is)

niksirbi (approved):
I'm approving this to unblock the other PRs, I only added minor suggestions for rewording.
In the future, maybe we can find a better solution than passing the `verbose` argument around across many functions?

JoeZiminski:
Thanks @niksirbi! Nice idea about the refactoring I've added to the validation issue.

PR #201, opened 2023-10-04
Perform check that we are running in TTY terminal before `getpass` is…

### LINKED ISSUE #179: [BUG] `getpass.getpass` hangs in some IDE on windows e.g. pycharm
problem is with `getpass.getpass` as described [here](https://stackoverflow.com/questions/24544353/python-getpass-getpass-function-call-hangs) [here](https://stackoverflow.com/questions/40007802/getpass-getpass-function-in-python-not-working)

One solution is to pre-check if this will happen and suggest to use the CLI.
Otherwise maybe a different package?

JoeZiminski:
closes #179

`getpass` fails in certain terminal environments that are not TTY terminals. See [here](https://unix.stackexchange.com/questions/4126/what-is-the-exact-difference-between-a-terminal-a-shell-a-tty-and-a-con) difference between all of these terminal types. I'm not 100% sure of the actual distrinction, but some consoles (e..g in the PyCharm IDE) are not true TTY terminals which results in problems reading the `stdin` and `stdout`, and `getpass` hangs.

The solution is to check we are running in a TTY terminal, if not throw an error rather than hang. I've tested this manually (not sure how to write an actual test for it, nor if it is needed). It errors running in PyCharm IDE (i.e. using Python API) but works fine using command-line-interface through terminal. I saw some edge cases online where this will fail in debug mode, wheras it should actually work in debug mode, but I don't think this case really concerns us, and it is best to be conservative.

### DIFF
--- a/datashuttle/utils/ssh.py
+++ b/datashuttle/utils/ssh.py
@@ -8,6 +8,7 @@ if TYPE_CHECKING:
 import fnmatch
 import getpass
 import stat
+import sys
 from pathlib import Path
 from typing import Any, List, Optional, Tuple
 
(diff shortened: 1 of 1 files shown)


niksirbi (approved):
WOW, this was great to learn. Solves some "mystery" problems I've had.

Issue #198, opened 2023-10-04
[Feature] Move to a new package the logic behind folder naming and validation


lauraporta:
**Is your feature request related to a problem? Please describe.**
Multiple projects need to infer folder names according to NeuroBluePrint specification for automated data processing and reading. 
Datashuttle currently owns part of the implementation required for the generation and the validation of these names.
However other projects are trying to handle the same technical necessities:
- [rsp_vision partially does this in the `SWC_blueprint` module](https://github.com/neuroinformatics-unit/rsp-vision/blob/main/rsp_vision/objects/SWC_Blueprint.py)
- [spikewrap validates folder names in its `base` module](https://github.com/neuroinformatics-unit/spikewrap/blob/dev/spikewrap/data_classes/base.py)

**Describe the solution you'd like**
As discussed in person, we could export this functionality in a standalone package.
The package should enable the user to:
- generate strings with folder names
- specify which set of key-value pairs we want in our session and subject folder
- validate these names
- expose objects containing the structure and the path to these folders
- being able to specify if we want raw/derivatives folders or ephys/anatomy/etc folders
- infer the next id number for session and subject
- handle cases of duplication
- having additional tools for batch analysis, as `analysied_files.log` and `exceptions_per_job_id.log`

**Describe alternatives you've considered**
An alternative could be importing datashuttle and using a dedicated module.

JoeZiminski:
Thanks Laura! I would propose the below plan:

1) Continue to list here and in #199 the required functionality as you have done
2) factor out / create all of this functionalty into datashuttle modules that implement this in functions.

try and use these from within datashuttle and see what the best approach is to structure it as a standalone module. There are a few possible options here as you note and it's hard to tell at this stage what will be the best pattern going forward but this is a good place to discuss.

(a bit of a tangent, looking forward)

For example, to perform validation (I think) it will always be necessary to hold the concept of the project root folder. We will also want a nice API to avoid making lots of calls to disperse functions, to which we will need to pass the project root. We will probably end up building a class that takes a project root and then exposes convenience functions like `project.make_sub_folder`. In this case `project` is an instantiation of a lightweight class that takes only a project root as input. But then this becomes quite close to what datashuttle already is, but with restricted functionality, so duplication and divergence may become an issue.

[...]

PR #203, opened 2023-10-09
[Validation] (2) Validation of passed `--sub_names` and `--ses_names`.

JoeZiminski:
This PR requires #206 as currently reserved keywords (e.g. `all_non_ses`) are also checked, leading to error.

Other PRs (#148, #204) that add validation check the names within a list against the rest of the entire project. However,
there was a blind spot in that sub / ses names within the list themselves were not checked, in particular
for duplicate ids or inconsistent leading zeros.

This PR adds this check at the level of making the subject directories. I don't think there is any other situation where
this check needs to be performed. Unless, it could be performed as an extra sanity-check when transferring data.

All this logic will be factored out at some point for #198 so maybe then is a better time to think about the structure.

### DIFF
--- a/datashuttle/configs/canonical_folders.py
+++ b/datashuttle/configs/canonical_folders.py
@@ -83,12 +83,12 @@ def get_non_ses_names():
     ]
 
 
-def get_keys_that_we_cant_format():
+def canonical_reserved_keywords():
     """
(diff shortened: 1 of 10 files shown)


niksirbi (approved):
Good to go other than 2 minor comments

PR #206, opened 2023-10-09
[Validation] (1) Refactor the code that performs validation of `--sub_names` and `--ses_names`.

JoeZiminski:
blocker for #203, which is rebased on this branch.

This PR does some minor refactoring to the logic that checks subject names and session names (e.g. that are passed to `make_sub_folders` or the transfer functions). 

1) rename `sub_or_ses_names` to `prefix` for consistency across functions
2) split `format_names` into `format_names` and `validate_names`. 

Currently, there are a list of `RESERVED_KEYWORDS` that can be passed to the transfer functions (e.g. `all_non_ses`) that we need to skip formatting and validation of. For formatting, these are skipped when required within the relevant function. For validation, all reserved keywords are stripped from the list before checking. 

There might be a better way to do this, because this is slightly not-robust for the formatting case. it would be possible if writing a new formatting fucntion to forget to check for reserved keywords. Also, the functions that format the tags do not check for the keywords, because they will never contain the tags and pass right through. Nonetheless, it is still weird to even pass the keywords to these functions.

Maybe it is better, similar to validation, to first split the passed list into `to_format` and `reserved_keywords` list. Then the `to_format` can be formatted, and the list rejoined at the end. This will be kind of ugly but I think much more robst.

### DIFF
--- a/datashuttle/configs/canonical_configs.py
+++ b/datashuttle/configs/canonical_configs.py
@@ -19,7 +19,7 @@ if TYPE_CHECKING:
 from pathlib import Path
 from typing import Literal, Union, get_args, get_origin
 
-from datashuttle.utils import folders, utils
+from datashuttle.utils import utils
 
 
 def get_canonical_config_dict() -> dict:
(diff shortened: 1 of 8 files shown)


niksirbi (approved):
I found no issues with this one.

Regarding this:
> Maybe it is better, similar to validation, to first split the passed list into to_format and reserved_keywords list. Then the to_format can be formatted, and the list rejoined at the end. This will be kind of ugly but I think much more robust.

you have thought about it for much longer than I have, but from my more distant vantage point this sounds like the more "logical" option

PR #210, opened 2023-10-10
[Testing] Refactor all relevant tests to use base fixture class.

### LINKED ISSUE #22: Find a neat way to avoid duplication of fixtures
Currently there is some duplication of fixtures across files (e.g. test_command_line_interface and test_filesystem_transfer). As far as I can tell any fixture with scope="function" (or I believe any scope parameter) must be defined in the file. But, would be good to double check this + look for pytest updates that provide any workarounds.

JoeZiminski:
closes #22.

This PR makes all relevant test classes inhereit from a base class that provides shared fixtures. For the most part, fixtures required provide a generic project with configs set or not set. Therefore fixtures that are exactly the same are re-used across projects. 

The two cases which this is not the case are SSH tests and logging tests, where the fixtures require extra logic. So, these test classes do not inherit from base class.

This PR also renames the projects so that `setup_project -> project` because in nearly all cases, the project from the fixture is setup. This means `project -> no_cfg_project`, which is only used in config tests.


niksirbi (approved):
LGTM

niksirbi (approved):
LGTM

PR #215, opened 2023-10-19
[TUI] Add DirectoryTree and Input widgets to Create Tab (2)

### LINKED ISSUE #185: Create empty main window with tabs (1)
The first step in the TUI is to create a main window app page that will hold a tab widget, with pages 'Create' and 'Transfer'. The aim is for switching tabs to switch the page in the area underneath (image below). It seems Textual has already combined Tabs and ContentSwitcher into a [TabbedContent](https://textual.textualize.io/widgets/tabbed_content/) widget which seems to be exactly what we want.

We can include simple text on each content window that just says 'create' and 'transfer' so we know the tab switching is working well.  Once this is done we can start building the 'Create' page.

Even at this early stage it is probably worth separating the css out into separate files (even if these are currently blank). We can follow the example at [ContentSwitcher](https://textual.textualize.io/widgets/content_switcher/#__tabbed_1_2) (where `CSS_PATH = "content_switcher.tcss"` and they show the example .css file). We can keep this blank or add one field to check it is working. If your happy with it, we can go with a structure such as:

```
.
└── datashuttle/
    ├── pyproject.toml
    ├── ....
    └── datashuttle/
        ├── configs
        ├── ...
        └── tui/
            ├── css/
            │   └── content_switcher.css
            └── view/
                └── app.py  # contains the app / content switcher
```
We can figure out exactly what to call these folders / files later when their function is clearer.

Image ( for now, lets ignore the 'Setup' tab, and the top bar with datashuttle and the minimize / restore / close buttons).
mainwindow 

To summarize, we can:

1) Create an initial app with TabbedContent widget with tabs 'Create' and 'Transfer'
2) add the text 'create' and 'transfer' to the content area, so that switching tabs switches the text.
3) create a `.css` file for the TabbedContent widget and add one field (e.g. the first feild from the example) to check it works.

Please let me know if there are any questions or anything is not clear!

### LINKED ISSUE #186: Add a file tree browser to the 'Create' tab (2)
requires #185 

This will entail adding the [DirectoryTree](https://textual.textualize.io/widgets/directory_tree/) to the 'Create' page. It should be aligned to the left of the page. We can start by it taking up to say roughly ~1/4 of the width of the page. We can remove the 'Create' text label that was used for testing in #185.

In terms of functionality, the two things we will want from this down the line are:

1) The top level folder (at least by default) should be the project name. It would be cool to just test this can be implemented by trying to configure it to default so that the top folder on the file tree is a specific folder. We can test this with a path on our individual local machines, then make a note of how this is done (if its possible). This will be difficult to test across PCs so we can just check if it's possible now, and if it's not make another issue.

2) We will (down the line) watch to fill the text input boxes with the folder names from the file tree brower. As a first instance, we can see if we can link a double-click event to folders on the directory tree. If a folder is double clicked, we can try to print the folder name to the console.

create_filedirectory

Overall we can:

1) remove the 'create' tab from #185 
2) Add a DirectoryTree widget to the left of the page (taking up say 1/4 of it)
3) See if we can configure it such that the top folder displayed is of our choosing (by suppllying a path)
4) See if we can configure it so that double-clicking the name of a folder prints it to console
5) Add a .css file for this widget

For now, let's contain all the functional logic (e.g. for printing) in the same file as the widget. Later we will probably factor this out.

b-peri:
## Description

This PR adds a functional directory-tree viewer and single user input field to the create tab. Double-clicking any directory on the DirectoryTree widget prints the directory path to the Input widget. The DirectoryTree root is currently set to the current User's home directory, but this can be easily adjusted by passing any absolute path (as a string) to the DirectoryTree widget as its first positional argument.

## References

Follows from PR #193, and closes #186.

## Checklist:

### DIFF
--- a/datashuttle/tui/app.py
+++ b/datashuttle/tui/app.py
@@ -1,8 +1,17 @@
 from pathlib import Path
+from time import monotonic
 
 from textual.app import App, ComposeResult
 from textual.binding import Binding
-from textual.widgets import Footer, Header, Label, TabbedContent, TabPane
+from textual.widgets import (
+    DirectoryTree,
+    Footer,
+    Header,
+    Input,
+    Label,
+    TabbedContent,
+    TabPane,
+)
 
 
 class TuiApp(App):
@@ -20,7 +29,7 @@ class TuiApp(App):
     TITLE = "DataShuttle"
 
     # Set path to TCSS
-    tui_path = Path(__file__).parents[0]
+    tui_path = Path(__file__).parent
     CSS_PATH = tui_path / "css" / "tab_content.tcss"
 
     # Set key-bindings
@@ -33,12 +42,31 @@ class TuiApp(App):
         yield Header()
         with TabbedContent():
             with TabPane("Create", id="create"):
-                yield Label("Create", id="create_text")
+                yield DirectoryTree(str(Path().home()), id="FileTree")
+                yield Label("Folder Name")
+                yield Input(
+                    placeholder="Double-click on any folder to fill this field.",
+                    id="subject",
+                )
             with TabPane("Transfer", id="transfer"):
                 yield Label("Transfer; Seems to work!")
         yield Footer()
 
+    prev_time = 0.0
+
+    def on_directory_tree_directory_selected(
+        self, event: DirectoryTree.DirectorySelected
+    ):
+        """
+        After double-clicking a directory within the directory-tree widget, replaces contents of the \'Subject\'
+        Input widget with directory name
+        """
+
+        click_time = monotonic()
+        if click_time - self.prev_time < 0.5:
+            self.query_one("#subject").value = str(event.path)
+        self.prev_time = click_time
+
 
 if __name__ == "__main__":
-    app = TuiApp()
-    app.run()
+    TuiApp().run()
--- a/datashuttle/tui/app.py
+++ b/datashuttle/tui/app.py
@@ -66,7 +66,7 @@ class TuiApp(App):
         """
 
         click_time = monotonic()
-        if click_time - self.prev_time < 0.5:
+        if click_time - self.prev_click_time < 0.5:
             self.query_one("#subject").value = str(event.path)
         self.prev_click_time = click_time
 
--- a/datashuttle/tui/app.py
+++ b/datashuttle/tui/app.py
@@ -28,17 +28,20 @@ class TuiApp(App):
 
     TITLE = "DataShuttle"
 
-    # Set path to TCSS
     tui_path = Path(__file__).parent
     CSS_PATH = tui_path / "css" / "tab_content.tcss"
 
-    # Set key-bindings
     BINDINGS = [
         Binding("ctrl+d", "toggle_dark", "Toggle Dark Mode", priority=True)
     ]
 
-    # Compose window
+    prev_click_time = 0.0
+
     def compose(self) -> ComposeResult:
+        """
+        Composes widgets to the TUI in the order specified.
+        """
+
         yield Header()
         with TabbedContent():
             with TabPane("Create", id="create"):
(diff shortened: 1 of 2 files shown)


datashuttle/tui/app.py
```diff
@@ -0,0 +1,72 @@
...
+    if __name__ == __main__:
+         app = MyApp()
+         app.run()
+
+    Initialises the TUI event loop and starts the application.
+    """
+
+    TITLE = "DataShuttle"
+
+    # Set path to TCSS
```
JoeZiminski, line 31: I would suggest removing this comment as from the nice variable names it is clear what is going on

datashuttle/tui/app.py
```diff
@@ -0,0 +1,72 @@
...
+    tui_path = Path(__file__).parent
+    CSS_PATH = tui_path / "css" / "tab_content.tcss"
+
+    # Set key-bindings
```
JoeZiminski, line 35: Same here, though I'm conflicted because the reader does have to do some inference from the `Binding()` initialisation  to determine it is key bindings specifically. Can the variable name be changed to `KEY_BINDINGS` or is this fixed in textual?

datashuttle/tui/app.py
```diff
@@ -0,0 +1,72 @@
...
+    BINDINGS = [
+        Binding("ctrl+d", "toggle_dark", "Toggle Dark Mode", priority=True)
+    ]
+
+    # Compose window
```
JoeZiminski, line 40: I would suggest moving this to a docstring on the `compose` function and expanding a little bit as to what this is doing i.e. I think composing the widgets (in the order specified) onto the TUI).

datashuttle/tui/app.py
```diff
@@ -0,0 +1,72 @@
...
+        yield Footer()
+
+    prev_time = 0.0
+
+    def on_directory_tree_directory_selected(
+        self, event: DirectoryTree.DirectorySelected
+    ):
+        """
+        After double-clicking a directory within the directory-tree widget, replaces contents of the \'Subject\'
+        Input widget with directory name
```
JoeZiminski, line 62:
```suggestion
```
combined with above suggestion

datashuttle/tui/app.py
```diff
@@ -0,0 +1,72 @@
...
+                yield Label("Transfer; Seems to work!")
+        yield Footer()
+
+    prev_time = 0.0
+
+    def on_directory_tree_directory_selected(
+        self, event: DirectoryTree.DirectorySelected
+    ):
+        """
+        After double-clicking a directory within the directory-tree widget, replaces contents of the \'Subject\'
```
JoeZiminski, line 61:
```suggestion
        After double-clicking a directory within the directory-tree 
        widget, replaces contents of the 'Subject' input widget 
        with directory name.
```

JoeZiminski: Just to keep within line limits (Black does not format strings unfortunately). I have previoulsy been referrring to the textual widgets with capital leter e.g. Input but, if you agree I think lowercase reads better in the end.

datashuttle/tui/app.py
```diff
@@ -0,0 +1,72 @@
...
+        Input widget with directory name
+        """
+
+        click_time = monotonic()
+        if click_time - self.prev_time < 0.5:
```
JoeZiminski, line 66: very nice, I would add to the docstring that this is the Windows default double-click time

datashuttle/tui/app.py
```diff
@@ -0,0 +1,72 @@
...
+                yield Label("Folder Name")
+                yield Input(
+                    placeholder="Double-click on any folder to fill this field.",
+                    id="subject",
+                )
+            with TabPane("Transfer", id="transfer"):
+                yield Label("Transfer; Seems to work!")
+        yield Footer()
+
+    prev_time = 0.0
```
JoeZiminski, line 55: I would suggest moving this to under the `BINDINGS` variable name above. Typically a class is initalised with an `__init__` call (e.g. below) that sets up the class constructer. In textual's case it does not use this pattern, but nonetheless I would move all initalised variables to the top to follow this convention that class attributes are defined at the top of the class.

```
class MyClass:
    def __init__(self):
        self.prev_time = 0.0
```

JoeZiminski, line 55: I would consider renaming this variable to `prev_click_time` just for explicitlness

JoeZiminski (approved):
Hey @b-peri great work this works like a dream, just some minor suggestions.


One consideration, with the double click, this also expands and hides the contents of the folder. I wonder if it is easier  to use a single right-click to fill the input field. I don't think it will be possible to stop the open/hide folder behaviour of the single left click in a way that feels natural. However, right-clicking also feels a bit weird.

I suggest we don't worry about this for now but maintan a list of things to ask for feedback on / suggest alternatives when we get some people to test, one of which could be this.

datashuttle/tui/app.py
b-peri: Unfortunately this variable name is fixed in textual!

JoeZiminski: Okay that makes sense, thanks!  We can leave the comment here

PR #227, opened 2023-10-20
Remove `use_` datatype e.g. `use_behav` configurations.

JoeZiminski:
This PR removes `use_behav` etc arguments from datashuttle and tests. This PR is required for #229.

Now, the user must specifically pass the datatype to make each time for the API / CLI. By default, `datatype=""` meaning so datatype folders will be created. If `all` is passed, then every possible datatype will be made.

This means the user will always have to re-type the datatype. However, with the TUI we can make this simple in the GUI, and in general the API is envisioned to be used for scripting, for which this will not be a problem.

### DIFF
--- a/docs/source/pages/documentation.md
+++ b/docs/source/pages/documentation.md
@@ -68,8 +68,6 @@ The command `make-config-file` is used for the initial setup of the project. The
 
 `connection_method`: `local_filesystem` or `ssh`. Local filesystem can be used if the *central* storage is mounted to the local machine. Otherwise `ssh` can be used.
 
-Finally, the *datatype* flags `--use_ephys`, `--use_funcimg`, `--use_histology`, `--use_behav` set the types of data required for the project on the local machine. While individual flags are optional, at least one must be chosen when initialising the project.
-
 ### Optional Arguments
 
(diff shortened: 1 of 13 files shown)


docs/source/pages/documentation.md
```diff
@@ -134,11 +132,10 @@ Another example call, which creates a range of subject and session folders, is s
...
 datashuttle \
 my_first_project \
-make-folders -sub 001@TO@003 -ses 010_@TIME@ -dt all
-```
 
-When the `all` argument is used for `--datatype` (`-dt`), the folders created depend on the *datatypes* specified during *configuration* setup. For example, if
-`--use_behav`, `--use_funcimg`, `--use_histology` were set during *configuration* setup, the folder tree from the above command (assuming the time is `4.02.48 PM`), would look like:
+make-sub-folders -sub 001@TO@003 -ses 010_@TIME@ -dt behav funcimg histology
+
+When the `all` argument is used for `--datatype` (`-dt`), the folders created depend on the *datatypes* specified during *configuration* setup.
```
niksirbi, line 138: I thought there are no datatypes specified in config any more? Should this go?

niksirbi (approved):
There is a lingering mention of the "use" flags in a documentation example (line 93 in documentation.md). 

Other than that, and the one comment I made, good to go.

PR #230, opened 2023-10-21
Remove the `.datashuttle_meta` folder that was created in datatype folders.

### LINKED ISSUE #228: remove `.datashuttle meta`
This is legacy from when tests needed a folder inside to transfer the folder. Now folders are never transferred if they are empty. Tests write a file before transfer.

JoeZiminski:
closes #228.

This PR removes the legacy `.datashuttle_meta` that were created in every datatype folder.


niksirbi (approved):
Good to go!

PR #233, opened 2023-10-23
[TUI] Implement Project Selection Splashscreen (4)

### LINKED ISSUE #217: Project Selection Splashscreen (4)
This will presumably be implemented after #187, but can easily be developed independently and combined with other elements of the TUI later.

When using DataShuttle, the user will first need to configure a new project, or select an existing one. To facilitate this interaction, we plan to implement a retro video game-style "Save Select" screen, as discussed here #184.

image

This will require the following features:
1. An additional named ```Screen``` object should be installed during App configuration and mounted on startup.
2. A New Project button located at the bottom of the screen (that initially does nothing).
3. Automatically detect all DataShuttle projects present on the local system using ```get_existing_project_paths_and_names()``` (implemented in PR #202) and dynamically yield extended ```Button``` widgets containing project name, date of last edit/action, and datatypes configured for each of these. If no projects are present on the system, only the New Project button will be shown.
4. Optionally, an additional ```.tcss``` file dedicated to this screen, in case a lot of styling is required.
5. If #187 has already been implemented, clicking on an existing project should install and display a new ```Screen``` object containing the `TabbedContent` widget (w/ `Create` and `Transfer` tabs) implemented in previous PRs. All functionality should now be preconfigured for the selected project.

b-peri:
## Description

This PR implements a new project selection splashscreen (`ProjectSelect`) for the DataShuttle TUI. This screen detects DataShuttle projects present on the local filesystem and and dynamically displays these as clickable buttons. Clicking on any project then yields a new `TabScreen` Screen object containing a `TabbedContent` widget (w/ Create and Transfer tabs) preconfigured for the selected project. The user can then easily return to the project selection splashscreen by clicking the `Return` button located at the top-right of the `TabbedContent` screen. `ProjectSelect` screen also features a dummy "New Projects" button, which will later be used to instantiate the "New project configuration" screen.

## References

Closes #217

## Checklist:

### DIFF
--- a/datashuttle/tui/app.py
+++ b/datashuttle/tui/app.py
@@ -4,6 +4,7 @@ from time import monotonic
 from textual.app import App, ComposeResult
 from textual.binding import Binding
 from textual.reactive import reactive
+from textual.screen import Screen
 from textual.widgets import (
     Button,
     Checkbox,
@@ -18,6 +19,35 @@ from textual.widgets import (
 )
 
 from datashuttle import DataShuttle
+from datashuttle.utils.folders import get_existing_project_paths_and_names
+
+
+class ProjectSelect(Screen):
+    """
+    This Screen contains DataShuttle's project selection splashscreen.
+    From here, the user can select an existing project or begin
+    initializing a new project.
+    """
+
+    TITLE = "Select Project"
+
+    def __init__(self):
+        super(ProjectSelect, self).__init__()
+        self.project_names = get_existing_project_paths_and_names()[0]
+
+    def compose(self):
+        yield Label("DataShuttle", id="main_title")
+        yield Label("Select project", id="name_label")
+        for name in self.project_names:
+            yield Button(name, id=name)
+        yield Button("New project", id="new_project")
+
+    def on_button_pressed(self, event: Button.Pressed):
+        if event.button.id == "new_project":
+            pass
+        else:
+            app.project = DataShuttle(str(event.button.id))
+            app.push_screen(TabScreen())
 
 
 class TypeBox(Static):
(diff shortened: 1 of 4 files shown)


JoeZiminski (approved):
Hey @b-peri looks great! This is really cool and the integration with Datashuttle functionality looks really nice. Down the line it may be easier to select projects with  a text  search-able dropdown selection widget, but this doesn’t seem to exist at the moment in textual (though see https://github.com/Textualize/textual/discussions/1378). I think some acquisition machines will have a lot of experiments on the same machine, so this could be useful to add a note on for.

I made a couple of suggestions / additions below:
1) I refactored slightly to make the project selector screen the base screen for the application. The reason for this is it will make the logic easier to follow as the GUI get more complex. For example, the main window raises a screen, then pops it to get back to the main window when the screen is finished, so you end up where you started. If the logic flow is more sequential (e.g. a screen raises a screen raises a screen) it may become difficult to track when things get bigger. 

One thing this means is that in some cases, the TguiApp class is passed to the created screen, and coordinating functions on this main class are called, for example `show_modal_error_dialog`. The same effect could also be achieved by using the `app` class, however in general if global variables can be avoided and the hierarchy made explicit it can be useful for tracking down bugs later.

2) Added a modal dialog to display errors rather and remove the temporary inputs previously used for this. If something crashes, an error box is raised (the `QuitScreen()` class. Textual is very cool!

3) Renamed the id's to a standard format we could try going forward - let me know what you think. The names are formatted like <window name>_<some description>_<widget type> (let me know if this does not make sense).

4) Removed the footer for aesthetic reasons (and we will have light / dark mode on an option menu somewhere)

5) Only fill the relevant box (sub- or ses- that was clicked). The reason for this is I believe a common use case will be copying over the formatting of a session from one subject to another. But we can keep this in mind for sure during user testing and let me know if it feels unintuitive.

6) I played around with the .tcss a lot to get a more standardised aesthetic, feel free to make any changes to this and highlight things you don’t like. Currently these changes make light-mode look awful but I will make a new set of stylesheets for light-mode.

Please let me know if you disagree with any of these changes or have any questions. Thanks again this is looking really nice!

PR #237, opened 2023-10-25
[TUI] Implement New Project Configuration Screen (6)

### LINKED ISSUE #218: New Project Configuration Screen (5)
Requires #217.

Once the project selection screen has been implemented, we can add functionality to the New Project button. This button should open a new screen containing a form in which all inputs required for `make_project_config` can be specified.

image

To resolve this issue, the following features should therefore be implemented:
1. Clicking the New Project button should mount a new, un-named `Screen` object containing `Input` fields and `Select` widgets as appropriate to provide the main inputs taken by `make_project_config`.
2. The new screen should feature a "Back" button which pops the new `Screen` object and returns to the Project selection screen.
3. At the bottom of the New Project Configuration screen should be a `Button` widget that runs `make_project_config` using the above inputs, pops the new `Screen` object, and switches to a `TabbedContent` screen preconfigured for the new project.

b-peri:
## Description

This PR adds a simple New Project Configuration form that can be accessed via the "Make new project" button on the newly-implemented main menu.

## References

Closes #218

## Checklist:

- [X] Tests have been added to cover all new functionality


JoeZiminski:
Hey @b-peri love this! really coming along now. I saw that textualize does have a way to disable widgets, so tried placing this screen as a new tab on the project page, would be good to get your feedback on this.

The idea to embed within a tab means this TabbedContent screen is the area of the GUI that deals explicitly with the project. Clicking 'Make New Project' takes you to the 'Config' tab with some info on what to do. At this stage 'Create' and 'Transfer' are disabled. If the project is sucessfully setup, they become usable (and the next time the Config tab has gone the setup-relevant widgets are removed).

Then, if the user wants to go back and edit the configs they can do so through the same panel. Currently the "ssh" option is disabled (will assert that the connection method is 'local_filesystem') and re-setting configs is not implemented. Also, I had to do some weird / messy things to get this to work in the first instance, so will tidy up (ideally before we meet tomorrow).

Some additional small thoughts:
- It might be worth using a [radiobutton](https://textual.textualize.io/widgets/radioset/), assuming this is the same as Qt it displays mutually exclusive options. The (only real) benefit of it here is it removes the need for the user to click an extra time to make the window drop-down.
- I think for now we can also include the transfer options here, and maybe move them later.
- To save vertical space, we could try inlining the label (e.g. Local Path) and the input.
- In it's pressent form, the error dialog is used to show a non-error message. We can think about if we want a general popup (grey) or like the red and want to split into error popup (red) and general popup (grey), or even red, grey, green depending on how easy it is to dynamically update the CSS from within the app (I think, you said this is quite easy?)

Thanks a lot things are really looking good!

JoeZiminski:
TODO: there is a bug if you 'Make New Project', set SSH to local filesystem and click 'Configure Project' then an error message is shown but the Create and Transfer tabs become un-disabled

- fill config widgets with the current configs of the project (for when not setting up as new project).

PR #247, opened 2023-10-30
Add more exception types (not just raising `BaseException`).

### LINKED ISSUE #246: Fix Error logic.
Currently to centralise errors all errors go through a central `log and raise error function`. Need to do a quick review of this but there doesn't seem to be any benefits of centralising the errors in this way. I thoguht it could be for the GUI but in the GUI case we just wrap datashuttle in an try / except and display the exception, so it is not useful.

It is particularly bad because all exceptiosn are BaseExceptions so for API users they cannot except specific exceptions.

JoeZiminski:
closes #246. It is useful to have a central function in which all errors are raised, for example if needing to tear down any functionality (e.g. close loggers) before raising an error.

Previously this centralisation meant that only `BaseException` was raised. Now, the central error-raising functions take the type of exception to raise, and raise this. These means API users can catch specific exception types.

This PR also introduces two custom exceptions, `NeuroBlueprintError` (related to violations of NeuroBlueprint standard) and `ConfigError` (related to user passing bad configs).

Updated relevant tests, docs not required.

### DIFF
--- a/datashuttle/utils/custom_exceptions.py
+++ b/datashuttle/utils/custom_exceptions.py
@@ -1,4 +1,4 @@
-class ConfigError(BaseException):
+class ConfigError(Exception):
     pass
 
 
--- a/datashuttle/utils/custom_exceptions.py
+++ b/datashuttle/utils/custom_exceptions.py
@@ -2,5 +2,5 @@ class ConfigError(Exception):
     pass
 
 
-class NeuroBlueprintError(BaseException):
+class NeuroBlueprintError(Exception):
     pass
--- a/datashuttle/utils/custom_exceptions.py
+++ b/datashuttle/utils/custom_exceptions.py
@@ -1,0 +1,6 @@
+class ConfigError(BaseException):
+    pass
+
+
+class NeuroBlueprintError(BaseException):
+    pass
--- a/datashuttle/configs/canonical_configs.py
+++ b/datashuttle/configs/canonical_configs.py
@@ -15,11 +15,11 @@ from typing import TYPE_CHECKING, List
 
 if TYPE_CHECKING:
     from datashuttle.configs.config_class import Configs
-
 from pathlib import Path
 from typing import Literal, Union, get_args, get_origin
 
-from datashuttle.utils import utils
+from datashuttle.utils import folders, utils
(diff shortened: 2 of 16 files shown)


datashuttle/utils/custom_exceptions.py
```diff
@@ -0,0 +1,6 @@
+class ConfigError(BaseException):
```
IgorTatarnikov, line 1:
```suggestion
class ConfigError(Exception):
```
Someone might try to catch the exception using:
```
try:
    something
except Exception as e:
    cleanup
```
And not catch the exception if it inherits from `BaseException`.

datashuttle/utils/custom_exceptions.py
```diff
@@ -0,0 +1,6 @@
...
+    pass
+
+
+class NeuroBlueprintError(BaseException):
```
IgorTatarnikov, line 5:
```suggestion
class NeuroBlueprintError(Exception):
```
Someone might try to catch the exception using:
```
try:
    something
except Exception as e:
    cleanup
```
And not catch the exception if it inherits from `BaseException`.

IgorTatarnikov (commented):
Looks great! My only comment relates to the custom exceptions inheriting from `BaseException`. Could that be changed to `Exception` instead without downstream consequences? 

Also I noticed that sometime you use relative imports instead of absolute, is there a benefit or is that just a style choice? e.g.
[https://github.com/neuroinformatics-unit/datashuttle/blob/570d4d6566e343b1d1a9ed0d85df24fd32a9acd6/datashuttle/utils/folders.py#L15](url)

JoeZiminski:
Thanks @IgorTatarnikov! That's much better to inherit from `Exception` cheers. 

Also good point about imports, I always thought they were preferred based on nothing, [I see](https://stackoverflow.com/questions/4209641/absolute-vs-explicit-relative-import-of-python-module) absolute is much preffered! which makes sense. On [a commit](https://stackoverflow.com/questions/4209641/absolute-vs-explicit-relative-import-of-python-module) in another refactoring PR I've changed all imports to absolute and added a `ruff` rule to error on relative imports. Thanks!

IgorTatarnikov (approved):
Looks good!

PR #250, opened 2023-10-31
Do not create folders above project name [1]

JoeZiminski:
The PR causes an error to be raised if the passed `local_path` does not exist. Unfortunately this ended up being slightly more complex than anticipated and leading to a few downstream effects, making this PR sad and bloated :(

This PR:
1)  Implements a `check_folder_above_project_name_exists()` function. This always checks if the `local_path` exists. If `connection_method=="local_filesystem"`, it will also check the `central_path` exists. However, it is not possible to check the `central_path` exists if SSH is used (because it's not guaranteed to be setup). So, if the `local_path` does not exist but connection method is SSH, then a reminder to check the `central_path` is shown, on the assumption the central path might also be wrong.
2) This change lead to a load of tests failing so have ensured that tests make all test paths themselves. Most of these changes can be ignored in review because they are critical to test components (i.e. the tests would fail if something was wrong with them).
3) A benefit of this change is it allows simplification of what was a super unclear behaviour when setting up logs. Previously, if the `local_path` was changed then the log would store in a temporary path and then move the logs to the new path once it had been created. Now, in the case of 'update-config' if `local_path` is changed the logs are just stored in the original `local_path` that definately exists (then, all future logs will be stored in the new `local_path`. In `supply-config-file` and `make-config-file` we cannot be sure the `local_path` exists so logs are stored in temp folder then moved. So the core behaviour is still there but it has been cleaned up.

Please let me know if anything is not clear because I still get confused by the logging behaviour (hopefully it will be clearer now with the changes, but probably difficult to review). This section is also not super critical to review, and behaviour is tested e.g. `test_logs_new_supply_config`. 

This behaviour is tested, I do not think it requires any docs.

### DIFF
--- a/datashuttle/configs/canonical_configs.py
+++ b/datashuttle/configs/canonical_configs.py
@@ -162,6 +162,14 @@ def check_dict_values_raise_on_fail(config_dict: Configs) -> None:
                 ConfigError,
             )
 
+    if not (config_dict["local_path"].parent.is_dir()):
+        utils.log_and_raise_error(
+            f"The local path {config_dict['local_path'].parent} that the project "
+            f"folder will reside in does not yet exist. Please ensure the  path shown "
+            f"in this message exists before continuing. Also make sure the "
+            f"`central_path`, is correct, as datashuttle cannot check it at this stage."
+        )
+
     # Check SSH settings
     if config_dict["connection_method"] == "ssh" and (
         not config_dict["central_host_id"]
--- a/datashuttle/configs/canonical_configs.py
+++ b/datashuttle/configs/canonical_configs.py
@@ -162,13 +162,7 @@ def check_dict_values_raise_on_fail(config_dict: Configs) -> None:
                 ConfigError,
             )
 
-    if not (config_dict["local_path"].parent.is_dir()):
-        utils.log_and_raise_error(
-            f"The local path {config_dict['local_path'].parent} that the project "
-            f"folder will reside in does not yet exist. Please ensure the  path shown "
-            f"in this message exists before continuing. Also make sure the "
-            f"`central_path`, is correct, as datashuttle cannot check it at this stage."
-        )
+    check_folder_above_project_name_exists(config_dict)
 
     # Check SSH settings
     if config_dict["connection_method"] == "ssh" and (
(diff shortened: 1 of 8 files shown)


tests/tests_integration/test_configs.py
```diff
@@ -77,6 +81,90 @@ def test_bad_paths(self, project, bad_pattern, path_type):
...
+        # exist, because it is not possible to check over SSH.
+        no_cfg_project.make_config_file(
+            existant_path / no_cfg_project.project_name,
+            non_existant_path / no_cfg_project.project_name,
+            "ssh",
+            central_host_id="fake_id",
+            central_host_username="fake_username",
+        )
+
+    def get_exists_and_does_not_exist_path(self, tmp_path):
```
niksirbi, line 158: Nitpicking here, but I wonder whtehr it's worth defining these two paths as fixtures. Alternatively, keep this functon as is but return a dict instead of tuple.

niksirbi: Also, I think the spelling should be "existent" (instead of "existant")

tests/tests_integration/test_configs.py
```diff
@@ -77,6 +81,90 @@ def test_bad_paths(self, project, bad_pattern, path_type):
...
+                central_host_username="fake_username",
+            )
+
+        assert (
+            "Also make sure the central_path`, is correct, as datashuttle "
+            "cannot check it via SSH at this stage." in str(e.value)
+        )
+
+        # This should not raise an error, even though the path does not
+        # exist, because it is not possible to check over SSH.
```
niksirbi, line 149: FYI you cand do:
```python
from contextlib import nullcontext as does_not_raise

with does_not_raise():
```

I don't know if it's useeful here, but it allows you to parametrize arguments against multiple expected exceptions, one of which could be `does_not_raise()` (in cases where some values raise exceptions and others don't)

niksirbi (approved):
Looks good, other than some small minot comments/questions I had.

datashuttle/configs/canonical_configs.py
```diff
@@ -196,6 +198,47 @@ def check_dict_values_raise_on_fail(config_dict: Configs) -> None:
         )
 
 
+def check_folder_above_project_name_exists(config_dict: Configs):
+    """
+    Throw an error if the path above the project root does not exist.
+    This validation is necessary (rather than simply creating the passed folders)
+    is to ensure the `local_path` or `central_path` are not accidentally set to a wrong
```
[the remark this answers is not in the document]
JoeZiminski: Thanks Niko! Fixed [here](https://github.com/neuroinformatics-unit/datashuttle/pull/286/commits/a3c25b10e0b7139469a5af01a361f910e1d9d1fc)

datashuttle/configs/canonical_configs.py
```diff
@@ -196,6 +198,47 @@ def check_dict_values_raise_on_fail(config_dict: Configs) -> None:
...
+    location.
+
+    If the `connection_method` is "ssh" is it not possible to check the central
```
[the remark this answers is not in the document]
JoeZiminski: Fixed [here](https://github.com/neuroinformatics-unit/datashuttle/pull/286/commits/a3c25b10e0b7139469a5af01a361f910e1d9d1fc)

datashuttle/configs/canonical_configs.py
```diff
@@ -196,6 +198,47 @@ def check_dict_values_raise_on_fail(config_dict: Configs) -> None:
...
+        return (
+            f"The {path_name}: {config_dict[path_name].parent} that the project folder "
+            f"will reside in does not yet exist. Please ensure the path shown in this "
+            f"message exists before continuing."
+        )
+
+    if not (config_dict["local_path"].parent.is_dir()):
+        if config_dict["connection_method"] == "ssh":
+            extra_warning = (
+                "Also make sure the central_path`, is correct, as datashuttle "
```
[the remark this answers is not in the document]
JoeZiminski: Fixed [here](https://github.com/neuroinformatics-unit/datashuttle/pull/286/commits/a3c25b10e0b7139469a5af01a361f910e1d9d1fc)

datashuttle/utils/utils.py
```diff
@@ -66,6 +67,7 @@ def raise_error(message: str, exception) -> None:
     """
     Centralized way to raise an error
     """
+    ds_logger.close_log_filehandler()
```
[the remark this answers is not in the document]
JoeZiminski: Created issue from above comment.

tests/tests_integration/test_configs.py
JoeZiminski: This is really cool, in this specific case I will leave it because the order of the arguments is swapped rather than parameterising a single argument so it might be a little fiddly. However will definitely remember this for future, thanks!

tests/tests_integration/test_configs.py
JoeZiminski: Thanks good idea! I think I originally did not do this because it needs `tmp_path` which is also  fixture; forgetting that fixtures can take fixtures as arguments. Fixed [here](https://github.com/neuroinformatics-unit/datashuttle/pull/286/commits/47da8d0eee9f4be390b4fee901f325e6d30bac66).

JoeZiminski:
Hey @niksirbi thanks for this! I have made relevant changes on #286 please feel free to leave any additional comments there. Cheers!

Issue #242, opened 2023-10-27
use `pytest` raises `match` argument.


JoeZiminski:
To avoid checking afterwards e.g. is `"str" in str(e.value)` or whatever it is currently.

Thanks @lochhh!

JoeZiminski:
In the end the regex nature of the search means special characters in the assert message [need to be escaped](https://stackoverflow.com/questions/55982221/unexpected-result-pytest-raises-with-match-parameter-when-asserting-error). As often paths are been tested and other longer messages including special characters, it is easier to continue to use the current form that does not reply on regex. However this is very cool and will generally use where posssible!

PR #251, opened 2023-10-31
Improve get next sub or ses funcs [2]

JoeZiminski:
The functions `get_next_sub_number` and `get_next_ses_number`  previously returned the next suggested number as an integer, as well as the maximum number. 

The functions work by checking the maximum number of all subject / sessions within the project (well, within the `local_path` and `central_path`). Now, only the next session number is returned (no real reason to return the maximum number), and is returned in a useful format, in BIDS style consistent with the number of leading digits the project has used, with our with the "sub-" or "ses-" prefix (depending on the settings.

This adds additional tests for this behaviour, no docs are needed as the docs are from the docstring.

The API is a bit weird here (e.g. there is `get_next_sub_number` and `show_next_sub_number` as `project` methods), this is discussed further in #256.

### DIFF
--- a/datashuttle/utils/folders.py
+++ b/datashuttle/utils/folders.py
@@ -576,8 +576,11 @@ def get_local_and_central_sub_or_ses_names(
 
 
 def get_next_sub_or_ses_number(
-    cfg: Configs, sub: Optional[str], search_str: str
-) -> Tuple[int, int]:
+    cfg: Configs,
+    sub: Optional[str],
+    search_str: str,
+    return_with_prefix: bool = True,
+) -> str:
     """
     Suggest the next available subject or session number. This function will
     search the local repository, and the central repository, for all subject
@@ -586,7 +589,6 @@ def get_next_sub_or_ses_number(
     pair values, and return the maximum value + 1 as the new number.
     A warning will be shown if the existing sub / session numbers are not
     consecutive.
-
 
     Parameters
     ----------
@@ -601,7 +603,6 @@ def get_next_sub_or_ses_number(
     latest_existing_num : the latest sub / ses number that currently
                           exists in the project.
     """
-
     if sub:
         bids_key = "ses"
     else:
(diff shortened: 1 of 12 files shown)


datashuttle/utils/folders.py
```diff
@@ -576,70 +584,156 @@ def get_local_and_central_sub_or_ses_names(
...
-    )
+        # First get the length of bids-key value across the project
+        # (e.g. sub-003 has three values).
+        all_num_value_digits = [len(value) for value in all_values_str]
 
-    if not utils.integers_are_consecutive(all_value_nums):
-        warnings.warn(
-            f"A subject number has been skipped, "
-            f"currently used subject numbers are: {all_value_nums}",
+        if not len(set(all_num_value_digits)) == 1:
```
niksirbi, line 715: sonar lint warns me to use the `!=` operator instead:
```suggestion
        if len(set(all_num_value_digits)) != 1:
```

tests/tests_unit/test_unit.py
```diff
@@ -294,16 +294,124 @@ def test_get_values_from_bids_formatted_name_error(self):
...
+        A warning should be shown in that the number of value digits are
+        inconsistent, as the user may be confused as to the real max.
+        """
+        bad_num_values_names = [
+            f"{prefix}-001",
+            f"{prefix}-02",
+            f"{prefix}-003",
+        ]
+
+        with pytest.raises(BaseException) as e:
```
niksirbi, line 339: you can use the "match" argument of `pytest.raises` to search for the expected error message.
But perhaps you want to do that across all tests in a future PR. It makes no actual difference, but it will save you some code lines.

niksirbi (approved):
LGTM!

tests/tests_unit/test_unit.py
JoeZiminski: Thanks it is a v cool argument I tried this in #242 but unfortunately it has some funny behaviour with regexp that made it more trouble that it's worth in few cases, unless I overlooked something..

datashuttle/utils/folders.py
JoeZiminski: Thanks fixed [here](https://github.com/neuroinformatics-unit/datashuttle/pull/286/commits/abac75adfe53ccd2148f3a89a6a8635e3b47e8d6).

PR #254, opened 2023-11-09
Whitelist links that linkcheck does not like.

JoeZiminski:
Ignore links that do not work with github actions link checking, see [here](https://github.com/neuroinformatics-unit/actions/pull/24#issue-1978966182)

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -80,6 +80,12 @@ html_show_sourcelink = False
 #  "**": [],
 #}
 
+# Ignore links that do not work with github actions link checking
+# https://github.com/neuroinformatics-unit/actions/pull/24#issue-1978966182
+linkcheck_ignore  = [
+    "https://neuroinformatics.zulipchat.com/#narrow/stream/405999-DataShuttle"
+]
+
 autodoc_default_options = {
     'members': True,
     "member-order": "bysource",
@@ -144,11 +150,6 @@ html_theme_options = {
  #   "page_sidebar_items": [],
 }
 
-# Ignore links that do not work with github actions link checking
-# https://github.com/neuroinformatics-unit/actions/pull/24#issue-1978966182
-linkcheck_anchors_ignore_for_url = [
-    "https://neuroinformatics.zulipchat.com/#narrow/stream/405999-DataShuttle",
-]
 
 # Redirect the webpage to another URL
 # Sphinx will create the appropriate CNAME file in the build directory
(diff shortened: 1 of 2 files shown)


docs/source/conf.py
```diff
@@ -80,6 +80,12 @@
 #  "**": [],
 #}
 
+# Ignore links that do not work with github actions link checking
+# https://github.com/neuroinformatics-unit/actions/pull/24#issue-1978966182
+linkcheck_ignore  = [
```
lochhh, line 85: I think you would still want to check that the base URL works, and only skip checking the anchors, rather than ignoring individual full URLs with `linkcheck_ignore`?: 
```suggestion
linkcheck_anchors_ignore_for_url = [
    "https://neuroinformatics.zulipchat.com/",
]
```

docs/source/conf.py
```diff
@@ -80,6 +80,12 @@
...
+    "https://neuroinformatics.zulipchat.com/#narrow/stream/405999-DataShuttle"
+]
```
lochhh, line 87: I think you would still want to check that the base URL works, and only skip checking the anchors, rather than ignoring individual full URLs with `linkcheck_ignore`?: 
```suggestion
linkcheck_anchors_ignore_for_url = [
    "https://neuroinformatics.zulipchat.com/"
]
```

niksirbi: +1. I think you need to add the base URL in `linkcheck_anchors_ignore_for_url`, as in @lochhh's suggestion. In your earlier commit, you had added the full path to the zulip channel, which wouldn't work I think.

JoeZiminski: Thanks both, I thought I tried this locally and it still errored, but maybe I had not organised it in this way. Will check on the CI, cheers!

JoeZiminski: Hey @lochhh @niksirbi indeed for some reason this does not work, do you have any idea why? seems like it should 🤔

lochhh: from the error output, it seems like sphinx is outdated: sphinx-5.3.0 
`linkcheck_anchors_ignore_for_url` is [new in version 7.1](https://www.sphinx-doc.org/en/master/usage/configuration.html#:~:text=tuple%20().-,New%20in%20version%207.1.,-linkcheck_auth)

JoeZiminski: Awesome thanks @lochhh that did it!! 🙏 🙏

niksirbi (approved):

PR #257, opened 2023-11-10
Add unit test for `test_get_values_from_bids_formatted_name()`.

JoeZiminski:
Adds a unit test for the core function `get_values_from_bids_formatted_name()`. This function extracts values from list of BIDS-like next `["sub-001", "ses-002"]` would return `["001", "002"]` (if `"sub"` key was passed). Some small fixes to the function itself asn related tests.

I convered some different cases but I wonder if there are any edge cases I'm not thinking of.

No documentation or tests are needed.

### DIFF
--- a/tests/tests_unit/test_unit.py
+++ b/tests/tests_unit/test_unit.py
@@ -238,7 +238,7 @@ class TestUnit:
         self, key, return_as_int, sort
     ):
         """
-        Unit test the function `test_get_values_from_bids_formatted_name()`
+        Unit test the function `get_values_from_bids_formatted_name()`
         which extracts values from BIDS-like names with a range of
         possible inputs.
         """
--- a/tests/tests_unit/test_unit.py
+++ b/tests/tests_unit/test_unit.py
@@ -283,7 +283,7 @@ class TestUnit:
             )
 
         assert (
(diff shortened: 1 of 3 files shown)


tests/tests_unit/test_unit.py
```diff
@@ -231,7 +231,76 @@ def test_duplicate_ids_in_list_of_names(self, prefix):
...
+        Check errors that catch unpredictable behaviour
+        are displayed correctly.
+        """
+        with pytest.raises(BaseException) as e:
+            utils.get_values_from_bids_formatted_name(
+                ["sub-001_date-12345", "sub-001_date-12345_date-23456"], "date"
+            )
+
+        assert (
+            "There is more than instance of date in"
```
b-peri, line 286: Is this a typo (i.e. did you not mean to write "There is more than _one_ instance of date in")?

b-peri (approved):
Hey @JoeZiminski, I noted some (potential) typos, but otherwise looks good!

tests/tests_unit/test_unit.py
JoeZiminski, line 286:
```suggestion
            "There is more than one instance of date in"
```

PR #264, opened 2023-11-13
General codebase refactor. [5]

JoeZiminski:
This PR performs lots of little refactorings. The only things of substance are:

1) Move anything that gets things about the project (e.g. next sub or ses number) or a `getters.py` module
2) Add the new `local_path`  to `[get_next_sub_or_ses_number](https://github.com/neuroinformatics-unit/datashuttle/pull/264/commits/eeb689c018bca72367c0fbfa2783c47fad2cbbbd)() which will only search local path for next sub / ses.
3) add tests for the above change.

The rest is just adding type hints, line length changes etc.

The `utils` module is too big now and needs splitting up, but I will leave fro a future PR.

Relevant tests added and no docs required.


datashuttle/configs/canonical_configs.py
```diff
@@ -18,28 +18,27 @@
...
-        "local_path": None,
-        "central_path": None,
-        "connection_method": None,
-        "central_host_id": None,
-        "central_host_username": None,
-        "overwrite_old_files": None,
-        "transfer_verbosity": None,
-        "show_transfer_progress": None,
+    canonical_configs = {
+        "local_path": Union[str, Path],
```
niksirbi, line 31: since you use this type a lot, you may benefit from defining a custom type alias:

```python
StrOrPath = Union[str, Path]
```
for Python >= 3.10:
```python
StrOrPath: TypeAlias = Union[str, Path]
```

In this particular example, probably it doesn't save you much space, but keep it in mind if the need arises.

datashuttle/configs/canonical_configs.py
```diff
@@ -18,28 +18,27 @@
...
+        "central_path": Union[str, Path],
+        "connection_method": Literal["ssh", "local_filesystem"],
+        "central_host_id": Union[str, None],
```
niksirbi, line 34: Perhaps more readable to define it as `Optional[str]`? But this might be my personal preference.

niksirbi (approved):
LGTM!

datashuttle/configs/canonical_configs.py
JoeZiminski: Thanks! No that's much better, an oversight!

datashuttle/configs/canonical_configs.py
JoeZiminski: Thanks! I will keep this in mind.

JoeZiminski:
Thanks Niko! Fixed [here](https://github.com/neuroinformatics-unit/datashuttle/pull/286/commits/7084352da667bb38c75bd759e058b914ec3abaec)

PR #266, opened 2023-11-13
Add zulip chat icon to website.

JoeZiminski:
Add the link to Datashuttle's zulip chat in the top-right corner of the website.


alessandrofelder (approved):
Haven't tested locally, but checked that it matches what BrainGlobe does, so :+1:

PR #268, opened 2023-11-13
[TUI] Major Refactor of TUI Code

### LINKED ISSUE #262: [TUI] Refactor TUI Code (6)
The TUI code is starting to become a bit bloated, and it is now becoming difficult to work on the TUI in parallel. To ameliorate these issues, we should perform a major refactoring of the TUI code into separate modules. In the General Meeting Notes issue (#184), I proposed the following re-organization:

image

- `tui_app.py`: The main Textual app for the TUI (`TuiApp`), and the script that the user will call from the terminal (using textual) when they want to run the TUI.
- `project_config.py`: `ConfigsContent`, `NewProjectScreen`, and associated objects involving configuration of a new project or updating configs for a pre-existing project.
- `tui_logic.py`: TUI-specific functions and logic
- `custom_widgets.py`: Custom widgets and **multi-use** modal windows built for the TUI (e.g. `DatatypeCheckboxes` or `ErrorScreen`, but not `ShowConfigsDialog`)
- TabScreen Tabs will be subdivided into:
     - `main_tab.py`: Containing the main `TabScreen` and associated (non-tab-specific) logic
     - `create_tab.py`: Create tab and associated logic
     - `transfer_tab.py`: Transfer tab and associated logic
     - And so on...
     - **NOTE:** The configs tab will likely not require a separate script from `project_config.py` 

However, this is by no means set in stone, so please let me know if a different organization would make more sense!

b-peri:
## Description

In this PR, we implement a major refactoring of `app.py`, as described in #262.

## References

Closes #262

## Checklist:


JoeZiminski (changes requested):
Hey @b-peri, looks great! Everything is nicely compartmentalised into logical groupings and this should make working on the code separately much easier in future.

All the classes are very easy to read, the only very long one is the `project_configs` class. At some point, we will aim to completely split all logic away from the 'raw' TUI code that handles  the actual widget / screen objects. We can basically go through everything and pull blocks of code  calling `DataShuttle` into a new module that the tui-objects call. 

But I think we are not there yet and this can be done right at the end all in one go. I think that will help reduce the size of the configs class so we can leave for now. 

Please see some suggestions below let me know what you think and if you disagree with any renamings, I am not 100% on these suggestions:

Overeall the names for the various modules / main classes could be:

```
tui.app.py, TuiApp
screens.project_manager.py, ProjectManagerScreen
screens.configs.py, ConfigsScreen
tabs.create_folders.py, CreateFoldersTab
tabs.transfer.py, TransferTab
screens.project_selector.py, ProjectSelectorScreen
```

and organisation:

```
.
└── tui/
    ├── app.py
    ├── utils/utils.py
    ├── screens/
    │   ├── project_manager.py
    │   ├── configs.py
    │   ├── modal_dialogs.py
    │   └── project_selector.py
    ├── tabs/
    │   ├── create_folders.py
    │   └── transfer.py
    └── custom_widgets.py  # does not include modal window anymore, move to modal_dialogs.py
```

details:

- [x] Added a 'screens' folder to keep screen objects in. Eventually we could have 'screens', 'tabs', 'widgets' as top-level 'raw' tui folders

- [ ] Rename `tui_logic.py`  to `utils.py`. I removed 'logic' in case it becomes confusing down the line if we introduce a `DataShuttle` layer module which will also perform logic. Currently this `utils` is a place for helping getter functions etc but maybe it's scope will change in future.

- [ ] Rename `tab_screen.py` to `project_manager.py`. I basically wanted to describe the overall screens function, I'm not sure about the new name though.

- [ ] Rename `create_tab.py` to `create_folders.py` (we can also rename `CreateTab` class to `CreateFoldersTab` class). This is not faithful to the exact name of the tab as appears to the user but I think it should be clearer when navigating the folders.

- [ ] Introduce a `modal_dialogs.py` file that holds the custom modal dialogs (can move `ErrorScreen` here from `custom_widgets.py`).

- [ ] Rename `project_configs.py` to `configs.py` as I think this is clear now that it resides in the 'screen' directory but let me know if not.

- [ ] Rename `project_select.py` to `project_selector.py`

- [ ] Remove word 'tab' from `transfer_tab.py` and `create_tab.py` as they reside in 'tab' folder and `tui` from `tui_app.py`.

- [ ] In the `configs.py` file, the order could be changed a little, to have what is now `ConfigsContent` as the first clas sin the file, as it is the most important. Then above each other class we could have banners like
```
# --------------------------------------------------------------------------------
# Display Configs
# --------------------------------------------------------------------------------

class ShowConfigsDialog...
```
before each other class, just to help navigate the file. I like these banners for nagivating but they are not everyones cup of tea so don't worry if your not a fan we don't need them. Alternatively, we could move `ShowConfigsDialog()` to `modal_dialogs.py` and `NewProjectScreen()` to it's own python file in `screens`, I think I prefer this latter solution.

b-peri:
Hey @JoeZiminski,

Thanks a lot for the extensive feedback! I agree with (and have now implemented) pretty much all of your comments to a T.

My only reservation was on renaming `ConfigsContent` to `ConfigsScreen`. In this case, I think it's important for it to remain clear that `ConfigsContent` is, in fact, not a `Screen` object,  but rather a `Container` that can be implemented modularly alongside all kinds of other stuff on the same screen (e.g. a banner). Because of this, I think that `ConfigsContent` may be a slightly more informative name for the time being. Let me know if this makes sense!

JoeZiminski:
hey Bradon thanks for this! I agree, actually I completely forgot how this was organised 🤦‍♂️ I agree this is much clearer, and makes sense to store the various ways of loading the configs container in the configs folder. In this case it does make sense to have the screens first in the file as they are higher-level, and then the container that they call the underneath. Sorry this is the opposite of what I said before! Otherwise have approved, cheers!

JoeZiminski (approved):

b-peri:
> hey Bradon thanks for this! I agree, actually I completely forgot how this was organised 🤦‍♂️ I agree this is much clearer, and makes sense to store the various ways of loading the configs container in the configs folder. In this case it does make sense to have the screens first in the file as they are higher-level, and then the container that they call the underneath. Sorry this is the opposite of what I said before! Otherwise have approved, cheers!

Hey Joe sorry for the confusion here but `ConfigsContent` _does_ currently contain all of the main Configs widgets and logic so I'm in favor of keeping it at the top haha! If that works for you I'll merge!

JoeZiminski:
Hey Brandon sounds good! Typically I put the classes / functions in (rough) call order within a module i.e. here the Screens are the higher-level objects and they call / hold as an attribute the `ConfigsContainer`. However this is not any PEP8 and a roughtly-used personal convention so very happy to merge as-is. Thanks!

b-peri:
**Conclusion**: After some further discussion in private, @JoeZiminski and I decided it would be most clear to just refactor `configs.py` out into separate scripts (w/ `NewProjectScreen` -> `screens/new_project.py`, `ShowConfigsDialog` -> `modal_dialogs.py`, and `ConfigsContent` staying in `configs.py`, which is now no longer under `\screens\`)

PR #269, opened 2023-11-13
Add rclone check function.

JoeZiminski:
This PR wraps [Rclone's check](https://rclone.org/commands/rclone_check/) command. This is a nice command that checks a source directory (in our case `local_path` against a target directory (in our case `central_path`) and returns a list of all files and whether they are:

- same across source and target
- different across source and target
- found in source only
- found in target only

It outputs this in a relatively parseable way. This PR introduces:

1) function to directly wrap the `rclone check` call
2) function to parse the output into a dict of paths separated based on the above aases
3) testing function

This PR includes a test, and is internal so no docs required. This function is aimed to be used as part of the TUI filetree to indicate  differences between local and central filesystems - anything that will case this easier please let me know!

### DIFF
--- a/datashuttle/utils/rclone.py
+++ b/datashuttle/utils/rclone.py
@@ -139,6 +139,22 @@ def transfer_data(
     return output
 
 
+def perform_rclone_check(cfg: Configs) -> str:
+    """"""
+    local_filepath = cfg.get_base_folder("local").as_posix()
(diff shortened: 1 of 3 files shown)


b-peri (approved):
Hey @JoeZiminski, this looks great! Happy to approve as is.

JoeZiminski:
Thanks @b-peri!

Issue #267, opened 2023-11-13
[CLI] Determine future of CLI


JoeZiminski:
Currently the CLI basically wraps the API function-for-function. However, there is quite a lot of overhead with this, and it is not very clear for the user. Also, currently for some cases (e.g. `get_next_sub_number()`) not all underlying arguments are exposed. However, the CLI is fixed with a good test suite so should work okay for the time being, even if it is suboptimal.

With the TUI development, it is likely the CLI may not be used often. If this is the case, we can remove it, we can figure this out after proper release.

If it is continued, it should be overhauled. #125, #221 will need to be adressed. Related to #125, simply wrapping the API 1-1 isn't working and is becoming unweidly. If the CLI is not deleted, it should be designed as proper interface (i.e. by merging commands, then we can just have a configs, upload, download showers method) and remove attempt to 1:1 correspondance with API.

JoeZiminski:
With the removal of working top-level-folder in #344, many of the CLI functions were broken. It is not worth the time to fix these now. There is a `NotImplementedError` on the main function (except for the TUI). I would have liked to keep the original code underneath but SonarCloud was failing the linting because of it, so will paste here and add a note that this original code can be found prior to #344.

I am leaning towards removing all CLI code except for TUI launching, as it is undocumented and confusing if anyone finds. Then we can add it back in if able to properly support / it is required a lot.
```
        if "func" in args and str(args.func.__name__) == "make_config_file":
            warn = "ignore"
        else:
            warn = "default"

        warnings.filterwarnings(warn)  # type: ignore
        project = DataShuttle(args.project_name, print_startup_message=False)
        warnings.filterwarnings("default")

        if len(vars(args)) > 1:
            args.func(project, args)
        else:
            utils.print_message_to_user(
                f"Datashuttle project: {args.project_name}. "
                f"Add additional commands, see --help for details"
            )
```

JoeZiminski:
The CLI code that is defunct was removed in #347 to avoid having unused code in the publicly available codebase. However this PR is there in case we decide to add it back.

JoeZiminski:
See relevant (currently closed, can reopen if using CLI) issues:
#125 
#221

JoeZiminski:
CLI defined superseded by TUI

PR #271, opened 2023-11-14
Change `DataShuttle` public 'shower' functions to getters. [6]

JoeZiminski:
This PR changes at lot of the 'showers' functions in the API to getters, which makes more sense from an API-focused perspective. Now the CLI functions print the output in the `command_line_interface.py` script.

Also, all CLI functions are now tested (mostly these showers were not tested previously).

In some cases, showing configs and local-file-tree are still exposed as these perform some `__repr__` like functionality. Otherwise everything is a getter.

Currently, the CLI is a bit weird (e.g. I'm not sure if these should be called 'getters' in the CLI as they can't return anything) and not all underlying arguments are explosed. As descripted in #267, if we proceed with CLI then CLI can be properly designed to break the link between CLI and API. For now though the CLI should be fairly well fixed (in a sub-optimal state) and we can decide what to do with it later.

Slight change to docs to reflect new function name, but for most part these are build automatically from the API / CLI.

### DIFF
--- a/datashuttle/command_line_interface.py
+++ b/datashuttle/command_line_interface.py
@@ -193,7 +193,7 @@ def make_folders(project: DataShuttle, args: Any) -> None:
     """"""
     kwargs = make_kwargs(args)
 
-    filtered_kwargs = {k: v for k, v in kwargs.items() if v is not None}
+    filtered_kwargs = remove_nonetype_entries(kwargs)
 
     run_command(project, project.make_folders, **filtered_kwargs)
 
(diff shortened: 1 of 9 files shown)


niksirbi (approved):
Looks good, I tried all these functions from the CLI as a user, and they work.

JoeZiminski:
Cheers! Fixed [here](https://github.com/neuroinformatics-unit/datashuttle/pull/286/commits/dbefe9f108c0b860b0dcb7e0cb150f8766f09abb)

PR #273, opened 2023-11-14
Small general refactorings and documentation. [7]

JoeZiminski:
This PR performs some additional refactorings and adds soem documentation. The only things of note are:

- add a cool Ruff argument to not allow relative imports (based on criticism of relative imports [here])(https://stackoverflow.com/questions/4209641/absolute-vs-explicit-relative-import-of-python-module).
- Add some documentation to data transfer arguments which was missing.


niksirbi (approved):
That's a useful `ruff` rule indeed! LGTM

niksirbi (approved):
That's a useful `ruff` rule indeed! LGTM

Issue #265, opened 2023-11-13
[TUI] Add Transfer Tab Functionality to TUI


b-peri:
(Requires #262)

The TUI is now missing one last key function: Data transfer. This will be implemented as an element of `TabScreen`. In the General Meeting Notes issue (#184), I proposed the following general outline for this new tab (based heavily on an earlier design by @JoeZiminski, shown in the same issue): 

<img width="582" alt="image" src="https://github.com/neuroinformatics-unit/datashuttle/assets/77279592/778c0aeb-401a-476f-825a-45b8fc222583">

In brief, the transfer tab should be populated with the following elements:
- A `DirectoryTree` widget displaying the local project files, and, if possible, their transfer status.
- `RadioButton` or `Select` widget where the user can select between three main transfer options: 1. All (Where the entire project is transferred), 2. Top Level (where only files contained within a user-defined Top-Level Folder are transferred), or 3. Custom (where the user defines more specific transfer parameters themselves using e.g. convenience tags).
- A dynamically rendered content box (or tab) where the user can input specific additional arguments required by each of the above transfer options.
- A `Transfer` button which takes the provided inputs and passes these to the correct DataShuttle command

PR #276, opened 2023-11-15
[TUI] Add validation to create folders

JoeZiminski:
This PR makes improvement to the 'Create' tab on the project manager screen that is used to create folders. It adds:

- Online validation of the subject / sessions that are input into the "Subject" and "Session" input widgets. Under the hood these call datashuttle validation functions.
- A new 'template settings' popup window that allows setting regexp options used as sub / ses name 'templates' that validation can also (optionally) be performed against. These are based on the changes made in #278 
- Double-left clikcing on an input widget will auto-fill with the current template, if it is not set it will just fill with the prefix e.g. `sub-`.
- Double right-clicking on an input will auto-fill with the next suggested sub / ses number. If templates is on, it will replace the relevant key-value pair with the suggested key-value.
- Note all validation and next sub / ses suggestions are currently done based on the local folder only.
As part of the above two general features atr introduced:
- ClickableInput, a Input where clicks emit a signal as suggested by Textual [here](https://github.com/Textualize/textual/discussions/3680)
- a new decorator `require_double_click`. This will run the function it decorates if a double-click signal is detected, otherwise nothing will happen (the logic is the same as previous implementation only generalised). 

There are no tests except for manual testing. There is no TUI documentation at the moment.


b-peri (approved):
Hey Joe, thanks a lot for this very extensive (and very cool) PR! I've added a few additional commits to clean up some redundant code in the `.tcss` files and `template_settings.py`, but otherwise this seems great as is!

One note: As you likely already noticed yourself, `create_folders.py` is starting to get a bit bloated (and hard to read!), so I think we'll need to come back and refactor out some of the auxiliary logic down the line (esp. things like validators). We can meet to discuss this once #265 is done and dusted if that works?

PR #277, opened 2023-11-15
Remove unnecessary system calls.

JoeZiminski:
It turns out system calls to `rclone` are quite slow, meaning sometimes initialising the project is slow because rclone call to check for rclone install is made every time datashuttle is initialised. Also, for large projects logging the whole file tree on every `make_folders` call is slow.

For testing, whenever a config file is setup this entails calls to `_setup_rclone_central_local_filesystem_config` which is slow. I thought monkeypatching this in tests would lead to huge speed increase and also break the test, but it did neither.

This can become a prototyping PR then:

Move to new PR:

1) move the rclone only show install error message when system call fails to a new PR
2) remove the file tree logging in `make_folders` in the same PR
3) definitely move rclone config setup to `_set_attributes_after_config_load` as is done here as we always want to do this for new configs

Investigate further:
- Why are the tests not failing when monkeypatching local filesystem rclone config setup?

### DIFF
--- a/datashuttle/datashuttle.py
+++ b/datashuttle/datashuttle.py
@@ -28,6 +28,7 @@ from datashuttle.utils.custom_exceptions import (
 from datashuttle.utils.data_transfer import TransferData
 from datashuttle.utils.decorators import (  # noqa
     check_configs_set,
+    check_configs_set_for_transfer,
     requires_ssh_configs,
 )
 
@@ -107,8 +108,6 @@ class DataShuttle:
         if print_startup_message:
             if self.cfg:
                 self.show_top_level_folder()
-
-        rclone.prompt_rclone_download_if_does_not_exist()
 
     def _set_attributes_after_config_load(self) -> None:
         """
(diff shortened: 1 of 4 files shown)


JoeZiminski:
See the top answer here, instead it might be easier to keep the subprocess running.

https://stackoverflow.com/questions/19102166/python-speed-command-line-call-via-subprocess-or-another-method

JoeZiminski:
Yes instead try holding an open process (subprocess.popen) and calling when needed with communicate()

JoeZiminski:
Okay on final consideration, don't use `Popen` because while it will work well during the lifetime of a `DataShuttle` object, closing the process when the class is destroyed is not simple (e.g. overriding `__del__` does not work reliably). Therefore, just limit `subprocess` calls as much as possible rather than worry about managing processes.

JoeZiminski:
superceeded by [3f159e0](https://github.com/neuroinformatics-unit/datashuttle/pull/314/commits/3f159e01826b3a8a82c8507ef6ff0b366f2f7195)

PR #278, opened 2023-11-16
Add name templates and improve validation

JoeZiminski:
This PR makes some changes to the validation logic of datashuttle in order to allow the TUI to operate properly / fix any problems discovered when working on the TUI. Also, it introduces `name_templates`.

`name_templates` are regexp strings that sub / ses folder names can be validated against during validation and making folders. DataShuttle now has a setter that update a entry in the `persistent_settings` dictionary with new `name_templates` and a getter that returns them. These are called within the TUI code.

Otherwise, this PR includes some general cleanups and minor fixes to the validation code:
- stores all `persistent_settings` defaults in `canonical_configs.py`.
- imrpoves the `validation.py` error messages to account for if a single or multiple sub / ses names after invalid.

The behaviour is tested. Currently it is not documented, with issue #292 opened for this purpose. The reason it is not documented at present is there is a still a bit of work to do to decide on the final form on the regexp. We will probably go for a simpler regexp, for sure at the TUI level. At the python API level it would probably make sense to take original regexps inputs.

### DIFF
--- a/datashuttle/utils/validation.py
+++ b/datashuttle/utils/validation.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, List, Literal, Optional, Tuple, Union
+import re
+from typing import TYPE_CHECKING, Dict, List, Literal, Optional, Tuple, Union
 
 if TYPE_CHECKING:
     from datashuttle.configs.config_class import Configs
@@ -21,6 +22,7 @@ def validate_list_of_names(
     prefix: Literal["sub", "ses"],
     error_or_warn: Literal["error", "warn"] = "error",
     check_duplicates: bool = True,
+    name_templates: Optional[Dict] = None,
     log: bool = True,
 ) -> None:
     """
@@ -47,6 +49,10 @@ def validate_list_of_names(
 
     log: bool
         If `True`, output will also be logged to "datashuttle" logger.
+
+    TODO: this is potentially slow because each function loops over
+    the whole list. I'd imagine it would be faster to loop once
+    over and pass each name to the sub-function individually.
     """
     if len(names_list) == 0:
         return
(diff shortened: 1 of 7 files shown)


datashuttle/utils/validation.py
```diff
@@ -47,6 +49,13 @@ def validate_list_of_names(
 
     log: bool
         If `True`, output will also be logged to "datashuttle" logger.
+
+    name_templates : Dict
+        e.g. {"name_templates": {"on": False, "sub": None, "ses": None}}
```
niksirbi, line 54: being pedantic here. It's better for the order of parameters in the docstring to match that in the functions signature (`log` comes before `name_templates` in docstring, but other way around in the signature).

datashuttle/utils/validation.py
```diff
@@ -66,6 +76,59 @@ def validate_list_of_names(
             raise_error_or_warn(message, error_or_warn, log)
 
 
+def names_dont_match_templates(
+    names_list: List[str],
+    prefix: Literal["sub", "ses"],
+    name_templates: Optional[Dict] = None,
+) -> Tuple[bool, str]:
+    """
+    Test a list of subject or session names against the respective `name_templates`,
```
niksirbi, line 85: This line is over the length limit, break earlier.

datashuttle/utils/validation.py
```diff
@@ -66,6 +76,59 @@ def validate_list_of_names(
...
+    a regexp template.
+
+    If checking `name_templates` is on, an invalid result will be given if the
+    name does not re.fullmatch the regexp.
+    """
+    if name_templates is None:
+        return False, "No `names_template` dictionary passed."
+
+    if name_templates["on"] is False:
+        return False, "Names templates is set to off and will not be checked."
```
niksirbi, line 95: typo? You meant "Name templates"

niksirbi (approved):
LGTM!

I'm a fan of name templates and the tidying up of all persistent settings together is satisfying

JoeZiminski:
Thanks @niksirbi! fixes made in [29677c3](https://github.com/neuroinformatics-unit/datashuttle/pull/286/commits/29677c37c071d942b0be18515dc876ced8fe7244)

PR #281, opened 2023-12-01
[TUI] Add Transfer Tab

### LINKED ISSUE #265: [TUI] Add Transfer Tab Functionality to TUI

b-peri:
Requires #276

## Description

This PR adds transfer functionality to the TUI's `ProjectManager` screen. As proposed previously (#265), this new functionality has been implemented using the following widgets:

- A new, custom `DirectoryTree` widget which dynamically displays the transfer status of files in the local project folder.
- A `RadioButton` widget through which the user can select their desired transfer mode (from "All", "Top Level", and "Custom")
- A "Parameters" `Container` which dynamically renders mode-specific input widgets according to the transfer mode selected above.
- A `Switch` widget that allows users to select whether they wish to _upload_ or _download_ their project data.
- A transfer `Button` which passes the provided inputs to the appropriate transfer function.
- A new `ConfirmScreen` modal dialog, which is displayed after the transfer button is pressed.
- An options `Button`, which currently does nothing.

## References

Closes #265

Like previous TUI PRs, this PR has only been tested manually. There is also not yet any TUI documentation.


b-peri:
While the core transfer functionality has now been implemented, there are a few features we will want to come back to in a later PR:
1. Most importantly, an `Options` screen, from which the user can toggle more advanced transfer settings.
2. Dynamic styling of `TransferStatusTree` for "Custom" transfers.
3. Dynamic rendering (and styling) of files contained only on **central**

datashuttle/tui/tabs/transfer.py
```diff
@@ -9,5 +33,352 @@ def __init__(self, mainwindow, project):
...
+            self.project.cfg.get_base_folder("local").as_posix() + "/", ""
+        )
+
+        if (
+            node_path.stem.startswith("sub-")
+            or node_path.stem.startswith("ses-")
+            or node_path.stem in get_datatypes()
+            # ) and not node.is_expanded:
+        ):
+            if any(node_relative_path in file for file in self.diff_paths):
```
JoeZiminski, line 366: for readability I would consider assigning this boolean to a variable name e.g.:

```
path_has_changed = any(node_relative_path in file for file in self.diff_paths)  #(if I understood it's function correctly)

if path_has_changed:
   ...
```

datashuttle/tui/tabs/transfer.py
```diff
@@ -9,5 +33,352 @@ def __init__(self, mainwindow, project):
...
+
+        text = Text.assemble(prefix, node_label)
+        return text
+
+    def format_transfer_label(self, node, node_label, node_path):
+        node_relative_path = node_path.as_posix().replace(
+            self.project.cfg.get_base_folder("local").as_posix() + "/", ""
+        )
+
+        if (
```
JoeZiminski, line 360: Could add quick comment before each section to explain the conditions as took a little time to figure out e.g.

`# set folders to orange if files within have changed`

datashuttle/tui/tabs/transfer.py
```diff
@@ -9,5 +33,352 @@ def __init__(self, mainwindow, project):
...
+
+        elif (
+            node_path.stem in get_top_level_folders()
+            # and not node.is_expanded
+            and self.diff_paths
+        ):
+            node_label.stylize_before("gold3")
+
+        else:
+            if node_relative_path in self.transfer_diffs["same"]:
```
JoeZiminski, line 377: nice, considering how complicated the logic is for these two functions in general the code is very clear and readable!

JoeZiminski (changes requested):
[EDITING] Hey Brandon this is looking great! I think this feels very intuitive to use and supports a lot of functionality.  It is really cool to press the button then see the data appear in the new folders! The code is very well written and can follow easily what is going on and the logic flow. Please see some suggestions on the code and some misc items below. 

1) I exposed some of the testing utils, see `tests/quick_make_project.py` which might come in handy for quickly making a toy project for manually testing. We can move this code out of the PR before merge.
2) It would be cool to add the double-click functionality from the create tab to the transfer tab e.g. on the 'Custom' select if you double click a sub it adds it to the input. It could add it as a list rather than replace the existing. This isn't high-priority for this PR though (and actually playing around with it the implementation for the create inputs is a little buggy / confusing). I think most people will transer everything at once in the first instance, so this could be an issue.
3) For large files or long transfers, we may need to freeze the TUI until transfer has completed and show a 'Transferring....' screen (even with dynamic update from `rclone`'s show-progress functionality, though I'm not sure how this would work in practice).  This is slightly higher priority because without it people might try and transfer the same files twice. Maybe for now 

For the create tab, I think moving the 'settings' button from the bottom makes sense as it was a weird place for it. I did like the idea though of having some separation beween the functional 'Make Project' button and auxillary 'Settings' button. In my head I had it as the region 'this is the doing region where I click to create folders'. I'm not sure if this is completely made up or is a useful distcintino to have, if so maybe the 'settings' button could be moved just to the right edge of the screen. Also, there is textual rendering issue (bug?) in which if the screen is very small so the overflow scroll bar appears, if you scroll to the bottom the buttons are not shown fully. We could create an issue for this, maybe it is outside of the scope of the PR.

Thanks again for this! Feel free to call any comment or suggestion out of scope for this PR and we can create an issue and tackle it later. There are a lot of things to consider for the transfer tab and it is quite complex (the implmentation is very neat considering the underlying complexity) so makes sense to split across a few PRs.

datashuttle/tui/tabs/transfer.py
```diff
@@ -9,5 +33,352 @@ def __init__(self, mainwindow, project):
...
+
+    def update_transfer_tree(self):
+        self.transfer_paths = self.get_transfer_paths()
+
+        transfer_tree = self.query_one("#transfer_directorytree")
+        transfer_tree.transfer_diffs = get_local_and_central_file_differences(
+            self.project.cfg
+        )
+        filtered_diffs = {
+            key: self.transfer_diffs[key]
```
JoeZiminski, line 275: should this be `transfer_tree.transfer_diffs[key]`?

datashuttle/tui/tabs/transfer.py
```diff
@@ -9,5 +33,352 @@ def __init__(self, mainwindow, project):
...
+            elif self.query_one("#transfer_toplevel_radiobutton").value:
+                if upload_selected:
+                    self.project.upload_all()
+                else:
+                    self.project.download_all()
+
+            elif self.query_one("#transfer_custom_radiobutton").value:
+                if upload_selected:
+                    self.project.upload(
+                        sub_names=self.query_one("#transfer_subject_input")
```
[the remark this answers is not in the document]
b-peri: Hi Joe, completely agree that doing this bit in two lines would have been more readable, but unfortunately I'm prevented from doing so by _Black_!

b-peri:
Hey Joe, thanks again for your extensive review! I've now implemented all of the short-term fixes we discussed during our meeting last week, so this PR should be good to go for now! I'll throw up separate issues for the remaining items.

b-peri:
Hi Joe, many thanks for your detailed review! I've now finished implementing all of the fixes that we will be including in this PR. I just had some small final notes:
- I've managed to fix the label format "overwriting" issue that we discussed previously (in which "unstaged" files were not formatting correctly if they had already been styled by `TransferStatusTree.format_transfer_label()`), so no need to worry about asking the Textual devs about this anymore!
- At the current moment it seems like `datashuttle.utils.rclone.get_local_and_central_file_differences` only returns diffs relative to the project's current top level folder. This means that it is not currently possible to display color-coded labels for files under `rawdata` _and_ `derivatives` simultaneously. I'm assuming users will be working in `rawdata` for the most part, so maybe this isn't a huge issue, but I think this would be worthwhile (and probably fairly straightforward) to expand the rclone wrapper function to return _all_ diffs (across both top-level folders) ahead of the initial release of the TUI/DataShuttle!

JoeZiminski:
Hey Brandon,

Thanks a lot for this, that's awesome about the styling! I added a couple of commits, many are prototype / suggestions so let me know what you think:

[`Try DIrectoryTree with grey guide styling`](https://github.com/neuroinformatics-unit/datashuttle/pull/281/commits/84ed5725c38e9c0c9f62993fde89ae6adbec15b2)

I tried in this PR to remove all the fancy styling Textualize does to the guide on the free as I found them distracting from the colouring on the transfer tab. The configs tab now looks more boring, but I think it is easier to read without all the random flashing colors. However, it does also look more boring. Also, not all cases are caught because I am still randomly getting orange and green flashes so will require a bit more work if we keep.

[``Add extra checkboxes to datatype transfer tab custom`](https://github.com/neuroinformatics-unit/datashuttle/pull/281/commits/69c5f6495e533ca977bb0a9c29da21fdd0ff7707)

This adds the 'all', 'all datatype' and 'all_ses_level_non_datatype'. These work functionallty but the labels are not currently rendered nicely, this can be adressed in #295 (possibly after this PR is merged).

[`Prototypy commit to extend diff colors to all top ...`](https://github.com/neuroinformatics-unit/datashuttle/pull/281/commits/3229917dd91dc5ad6ccfc860562c8db9244384d0)

This makes a lot of changes to both `transfer.py` and on the datashuttle-side code, so will need tidying up if we keep it. It addresses your second point and makes both rawdata and deriatives show changes (in "All" mode). In the other two modes only the diffs under the selected top level folder are shown. It also refactors `transfer.py` so that folders that have 'local_only' within turn green. Folders that have both changed and local_only files within default to orange. This will also work similarly for error (if we ever see such a case). 

There is quite a lot of looping here, this is quite fast for smaller projects but if a project have many folders and files it might get quite slow. I'm not certain whether if it is possible to do without so many loops or we can cache some values somewhere. But for now we can keep a note of it and address optimisation later once the need is more pressing and tests are written. 

The only things I can think of are:

1) The folders change their color when they are clicked on, I think due to the focus CSS. I can't figure out a way to override this, I asked on their discord:


```
Joe — Today at 15:04
Hi everyone, I have a question about styling the DirectoryTree. We are setting custom colors on the text of specific nodes with node.stylize_before (or node.styles.colr = "color" also works). However, when clicking the node the text changes back to it's original color.  I can customise this color by changing TCSS DirectoryTree:focus .tree--cursor  but I would like for the focus to not change the color at all. Is it possible to block this or set the TCSS such that the color doesn't change?

davep — Today at 16:22
To the best of my knowledge (and with a quick glance at the source for Tree), as it stands right now, you can't style the Tree cursor such that there is no styling and the underlying label's style is used.

Joe — Today at 16:49
Thanks a lot for your your help, that makes sense. Do you know if it is possible to subclass the focus-handling machinery for Tree and it stop it triggering a style refresh at all?

davep — Today at 17:33
I believe the code that handles that is within the line drawing code, so not easily overridden, as it stands (I’m not anywhere near a place to check the code right now; but that’s my recollection).
```

So this seems like it's going to be pain, unfortunately it is very distracting however. Maybe we can find a way around this if we need a deeper sublcassing of DirectoryTree when we try to deal with the refreshind behaviour.

2) I need to align the Create tab inputs to handle multiple inputs the same was as transfer does. I'm still amazed how well this works off the bat for taking multiple lists!

Overall it's really coming along, extremely satisfying to use and looks fantastic!

JoeZiminski:
Hey @b-peri sorry for the delay on this! I added a few new features and did some refactorings, please let me know what you think and if you disagree with any changes made. 

I think the main thing left to sort out is the refreshing of the tree such that the selected node stays open, for which there is that open PR on textual. I tried to use their fork and the feature worked nicely but was getting random async event errors. We can look into this more but will wait to hear back from Will on that thread before proceeding.

Some changes / new additions:

-	Added a number of new keyboard shortcuts when working with inputs / directorytree (both here and in create folders tab):
o	To fill an input from the DirectoryTree, now howevering over the DirectoryTree with the mouse (when DirectoryTree has focus) and pressing CTRL+F will ‘fill’ the name into the relevant input. Pressing CTRL+A will append the name to the relevant input as you had. This is instead of double clicking which was a bit more convenient but open / close the folder, I dont think there is anything we can do about that. it still feels fairly natural.
o	I tried watchdog and could get it working in a minimum example in a python kernel but could not get it to work in the textual context. I think this might be a problem with threading as watchdog spins up its own thread. This could be a rabbithole so went for a quicker solution, now the DirectoryTree updates when tab is changed and also when CTRL+R is pressed (when DirectoryTree has focus)
o	For Inputs and DirectoryTrees, now pressing CTRL+Q when the widget has focus will copy the input value / tree node path (the path that the mouse is hovered over). Unfortunately CTRL+C is reserved by textual for closing the program, I hope this is not too confusing for people. I tried to override textual’s events and use ESC for exit and CTRL+C for copy, but it still quit on pressing CTRL+C, I wonder if it also relates to the terminal itself. Anyways, did not look further into it for now, it causes problems down the line we can revisit. In a similar context, pressing CTRL+O will open the relevant filepath in the filesystem browser (only tested on windows). If the value of the input is not a path it will handle and show user a modal window.
-	Removed ‘not staged for transfer’ for now as discussed. I also removed ‘Options’ button for now as the configs can be moved here at a later date in the interest of time. 
-	Added extra datatype checkboxes for the transfer context, ‘All’, ‘All Datatype’ and ‘All Session Level Non Datatypes’. These names are way too long and confusing, but will address after merge with #295
-	Override textual’s Tree() _render_line function to not apply any CSS when a tree node is clicked to stop the color changing. This is subject to textual issue 4028, need to check in with them because they couldn’t reproduce this problem.
-	Added a ‘Top level folder’ select to the custom transfer settings too. Currently this is not available for Create tab but can add later as described in #298. The actual implementation of this in the transfer-related functions (see the TODO in the docstring) is kind of hacky to ensure the top level folder is reset to what it was before transfer, but potential solutions might also be confusing, will leave for now.
- 
I think once this is merged into tui-dev-branch are basically ready to merge tui-dev-branch! Still some small changes to me made to the TUI but it is basically done!!

PR #342, opened 2024-03-28
Deprecate 'supply_config_file'.

JoeZiminski:
As described in #331, removing `supply_configs` for now because of the maintaince burden and reduced use since configs have been trimmed down masssively. 

A source of potential bugs is the supply-config comes with a local-path from a different project and somehow overwrites data that exists on the original project. supply-config should probably not copy the local path or ensure the project doesn't already exists.

This is still a cool feature as people may want template configs to create projects from. In the first instance though it is probably not too much bother for people to re-enter the same info a couple of times. If this is a requested feature all the code to implement is here on the PR.


niksirbi (approved):
> A source of potential bugs is the supply-config comes with a local-path from a different project and somehow overwrites data that exists on the original project. 

Scary indeed! Hadn't thought of that.
LGTM.

Issue #340, opened 2024-03-28
Make bypass validation on `create_folders` an argument


JoeZiminski:
Currently this is strange persistent config because it needs to be for TUI.

Increase split this so `create_folders` takes it as a direct argument (default `False`) and this and top level folder are just handled
on the tui persistent settings side. This is much more natural linked to #338 

Once done update API docs to include these argmunets.

PR #343, opened 2024-03-28
Reconfigure bypass validation

JoeZiminski:
As raised in #340, the machinery to bypass validation is designed from the TUI-perspective and set up as a persistent setting on the `DataShuttle` class. From the API perspective this is quite strange and added to an accumulation of hidden states on the class that were confusing.

Because people are unlikely to be using the API interactively (use TUI for that) and mostly in scripts, it is okay to have the key functions take a few more arguments. Here `bypass_validation` is moved as a direct `create_folders` method argument. For the TUI, the setting is now stored in the `"tui"` field of the `persistent_settings` dict.


datashuttle/datashuttle.py
```diff
@@ -168,6 +168,7 @@ def create_folders(
         sub_names: Union[str, List[str]],
         ses_names: Optional[Union[str, List[str]]] = None,
         datatype: str = "",
+        bypass_validation: bool = False,
```
niksirbi, line 171: also needs to be added to the docstring.

niksirbi (approved):
I agree with the choice!

datashuttle/datashuttle.py
JoeZiminski: https://github.com/neuroinformatics-unit/datashuttle/pull/345#discussion_r1550613548

Issue #338, opened 2024-03-28
Should top-level folder be a required input?


JoeZiminski:
We currently have the concept of 'working top-level folder' in the API. This is to avoid the user having to type `rawdata` or `derivatives` every time they make a folder or do customt transfers.

However, when writing the docs there is a whole page now dedicated to this top-level folder behaviour. Now with the TUI, if
users want a conveinent want to transfer or create folders they will use the TUI which does not require this 'working top level folder'
concept.

For the Python API, it will almost always be used in solidified pipelines where adding an extra, explicit argument is not much of a pain, and I doubt rarely used on the fly in the python console.

Therefore I think it makes most sense to remove this concept (as well as the docs page I am writing) and force the top level folder to be specified in `make_folders`, `upload_data`, `download_data`, `upload_all`, `download_all`.

PR #344, opened 2024-03-28
Remove working top level folder

JoeZiminski:
As discussed in #338 and related to #340, the concept of persistent settings to avoid having to input arguments to the API repeatedly makes less sense now with the TUI. Rather than interactively using API or CLI, interactive use will be through TUI. As API will be mostly scripts we can be a bit more demanding on the required arguments for some public facing methods.

This PR removes the persistent working 'top_level_folder' setting and now requires this as direct input to `create_folders`, the transfer methods, and validation methods. In the end this results in much more explicit code and makes the user more aware of the top-level folder concept.

This requires quite extensive changes to the code but they are all quite basic conceptually. For the review it probably makes most sense just to give feedback on the concept and look at the new signatures of `create_folders` and `upload` or `download`.

This PR also breaks a lot of the CLI which requires `top_level_folder` arguments. I tried adding `NotImplementedError` to the relevant sections but SonarCloud didn't like it. For now comments indicate where the CLI is broken, and the CLI code is removed for now in #347.

### DIFF
--- a/datashuttle/datashuttle.py
+++ b/datashuttle/datashuttle.py
@@ -474,7 +474,7 @@ class DataShuttle:
 
     @check_configs_set
     def upload_specific_folder_or_file(
-        self, filepath: str, dry_run: bool = False
+        self, filepath: Union[str, Path], dry_run: bool = False
     ) -> None:
         """
         Upload a specific file or folder. If transferring
@@ -502,6 +502,9 @@ class DataShuttle:
             transfer was taking place, but no files will be moved. Useful
             to check which files will be moved on data transfer.
         """
+        if isinstance(filepath, str):
+            filepath = Path(filepath)
+
         self._start_log("upload-specific-folder-or-file", local_vars=locals())
 
         top_level_folder = self._get_top_level_folder_from_specific_filepath(
@@ -510,7 +513,7 @@ class DataShuttle:
 
         processed_filepath = utils.get_path_after_base_folder(
             self.cfg.get_base_folder("local", top_level_folder),
-            Path(filepath),
+            filepath,
         )
 
         include_list = [f"--include {processed_filepath.as_posix()}"]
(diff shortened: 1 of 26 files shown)


datashuttle/datashuttle.py
```diff
@@ -198,6 +161,10 @@ def create_folders(
                 NeuroBlueprint will be created. If "" is passed
                 no datatype will be created.
 
+        top_level_folder :
+                Whether to make the folders in `rawdata` or
+                `derivatives`.
```
niksirbi, line 166: As a rule, it's better for the order of params in the docstrign to match that in the function signature, so I would move this up as the first param.

datashuttle/datashuttle.py
```diff
@@ -553,21 +562,30 @@ def download_specific_folder_or_file(
...
+            filepath = Path(filepath)
+
         self._start_log(
             "download-specific-folder-or-file", local_vars=locals()
         )
 
-        self._display_top_level_folder()
+        top_level_folder = self._get_top_level_folder_from_specific_filepath(
+            filepath
+        )
```
niksirbi, line 574: In the docstring of `download_specific_folder_or_file` and `upload_specific_folder_or_file` you mention:
> filepath :
>   a string containing the filepath to move,
>   relative to the project folder "rawdata"
>   or "derivatives" path (depending on which is currently
>   set). Alternatively, the entire path is accepted.

This is no longer true, right? The docstring needs to be updated according to the new behaviour.

niksirbi (approved):
I think you missed updating some docstrings (see individual comments) but other than that, I wholly agree with the concept.

With the persistent setting, there was always the risk of the user forgetting the top level folder value and thus making unintented transfers.

Having that as a required argument in the API is a small price to pay to avoid such mistakes.

datashuttle/datashuttle.py
niksirbi: Nevermind, I saw you have fixed this as part of [another PR](https://github.com/neuroinformatics-unit/datashuttle/pull/348).

PR #345, opened 2024-03-29
Return paths from create folders

JoeZiminski:
This PR makes `create_folders` return the created folders. This should be useful for using datashuttle in scripts as you can create folders then grab the path for saving data too.

This is in the new docs #334 and added to tests in this PR.


datashuttle/datashuttle.py
```diff
@@ -165,6 +165,10 @@ def create_folders(
                 Whether to make the folders in `rawdata` or
                 `derivatives`.
 
+        bypass_validation :
```
JoeZiminski, line 168: Sorry this change is unrelated to the main thurst of this PR.

niksirbi (approved):
I tested this locally and it works as intended. LGTM!

PR #348, opened 2024-03-30
Adress sonarcloud issues

JoeZiminski:


pyproject.toml
```diff
@@ -97,14 +96,14 @@ line-length = 79
...
 
-[tool.ruff.per-file-ignores]
+[tool.ruff.lint.per-file-ignores]
 "__init__.py" = ["F401"]
 
-[tool.ruff.mccabe]
+[tool.ruff.lint.mccabe]
 max-complexity = 18
 
 [tool.ruff.lint.flake8-tidy-imports]
```
niksirbi, line 109: I might be mistaken, but I think you can have a `[tool.ruff.lint]` section, under which you can group all setting that start with `lint`.

niksirbi (approved):
LGTM!

pyproject.toml
JoeZiminski: Thanks I changed the main header to `[tool.ruff.lint]` and removed `lint` from the entries within it and it worked. It may be also possible to add the other sections like this also, but I could not see example in the docs e.g. for `mccabe`. But don't have time to test ATM but will keep an eye on this in future!

JoeZiminski: Oh wait I see what you mean now, thanks!

PR #349, opened 2024-03-31
Fix unspecific error when validating names

JoeZiminski:
This fixes an annoying issue with validation on the TUI in which if there was an invalid name in the project it would show a validation error on the subject / session input box. This is quite confusing at it was unrelated to the actual name in the box.

This PR skips checking against invalid names for duplicate subject / sessions. This means it will only error if there is a duplicate with a valid NeuroBlueprint name. This is fine as if they input an invalid NeuroBlueprint name, even if there is a duplicate it will just show a standard NeuroBlueprint error.

### DIFF
--- a/datashuttle/utils/validation.py
+++ b/datashuttle/utils/validation.py
@@ -412,11 +412,8 @@ def validate_names_against_project(
     check that these names are formatted consistently with the
     rest of the project.
 
-    For basic checks, the new subject / session names are concatenated
-    with the existing ones and checked for consistently in
-    `validate_list_of_names()`. Note this implicitly checks that the
-    passed names are consistent with each-other (i.e. within `sub_names`
-    and within `ses_names`).
+    The passed list of names is first validated in `validate_list_of_names()`
+    without reference to the existing project.
 
     Next, checks for duplicate subjects / sessions are performed. For subjects,
     duplicates are checked for project-wide. For sessions, duplicates are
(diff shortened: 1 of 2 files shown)


niksirbi:
FYI this was my valid tree before I added the invalid `ses-03-02` under `sub-01`

Screenshot 2024-04-04 at 18 20 48

JoeZiminski:
Thanks @niksirbi good spot! Some additional changes have been made, now the existing project names are not passed to initial validation at all. Must have completely missed this! I've tested and I think it's looking okay now but let me know how it works for you.

niksirbi (approved):
Thanks @JoeZiminski, it works now 🎉 . LGTM

JoeZiminski:
Hey @niksirbi thanks for this! in the end this was more fiddly that anticipated and required a further rewrite. For posterity I will detail below as its a bit fiddly and now the validation functions are a bit confusing, I'll open an issue on this.

There are three types of validation that checking a list of names against the project needs to consider:
1) validation errors within the passed list of names
2) validation errors within the existing project
3) 'interaction' validation errors that occur e.g. a name in the passed list of names that duplicates an existing project name or has inconsistent sub or ses value lengths with the rest of the project.

In the function we need to ignore 2) (validation errors within the existing project). However, we can't ignore when there are inconsistent sub or ses value lengths within the existing project because then we are not sure what length to validate the passed list of names against.

The pre-PR implementation failed at 2), raising errors when there were validation issues in the existing project
The next implementations failed to catch the 'interaction' error of inconsistent sub or ses value lengths.
This final implementation now also catches inconcistent sub or ses value lengths. In the case that there are already inconsistent value lengths in the existing project, and appropriate error is raised stating validation can't be performed. I think this is OK even though it is strictly an issue with the existing project not the list of names, because there's nothing to be done about it. People can always `bypass_validuation=True` if they need to make a folder urgently.

Tests are now added to cover all these cases.

PR #359, opened 2024-04-09
Rename `get_next_sub_number` and `get_next_ses_number`.

JoeZiminski:
Straight CTRL+R rename of `get_next_sub_number` and `get_next_ses_number` to `get_next_sub` and `get_next_ses` as these functions actually return by default the key-value pair (e.g. `sub-001`).

Performed on both `.py` files and `.md` files so should cover docs. Checked with hypen separators instead of underscore (but did not find any as we don't have CLI anymore).

### DIFF
--- a/datashuttle/datashuttle.py
+++ b/datashuttle/datashuttle.py
@@ -851,7 +851,7 @@ class DataShuttle:
         local_only: bool = False,
     ) -> str:
         """
-        Convenience function for get_next_sub_or_ses_number
+        Convenience function for get_next_sub_or_ses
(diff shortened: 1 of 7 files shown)


niksirbi (approved):
Approved!

Any particular reason you left `getters.get_next_sub_or_ses_number()` unchanged?

It could become `getters.get_next_sub_or_ses()` (unless this one returns the actual number)?

JoeZiminski:
No I completely forgot about this function / did not catch in searches 😆  thanks!!

PR #362, opened 2024-04-10
Extend overwrite option and improve docs

JoeZiminski:
This PR had one main objective but unfortunately also includes a few general refactorings and improvements to data transfer.

The main aim is to extend the use of `overwrite_existing_files` from a simple `bool`, which did not well capture Rclone default behaviour, to `never`, `always`, or `if_source_newer`. Previously the default Rclone ovewrite behaviour was thought to be only-if-the-source-is-newer. In fact it will overwrite on any difference in size or modification datetime. These new options better capture this behaviour. TUI, docs and tests updated also.

On the way it made happened that the `dry_run` option was extended to the TUI transfer tab. Test are updated too.


niksirbi (approved):
Looks good. I tested the new "overwrite" option locally on Mac, using either a mounted dropbox folder or a moundted ceph drive as remotes.

The behaviour seems to be as described. I haven't tested transferring via the ssh connection though.

JoeZiminski:
Thanks @niksirbi !!!

Issue #380, opened 2024-04-29
Add `PyDantic`-style validation to all public functions


JoeZiminski:
It would be nice to properly validate all inputs to public functions. This could be done following the [movement](https://github.com/neuroinformatics-unit/movement/blob/main/movement/io/validators.py) way.

Currently this is done ad-hoc in datashuttle e.g. #375. This should be reworked to be centralised in a argument-validation module.

niksirbi:
See https://github.com/neuroinformatics-unit/movement/issues/24 and the linked articles to make an informed decision on `PyDantic` vs `attrs`. They both do very similar things, but int he end I went with `attrs` with `movement`.

JoeZiminski:
Thanks @niksirbi I think it makes sense to follow your lead, also good for consistency across NIU repos.

PR #381, opened 2024-04-29
Add _check_top_level_folder option.

JoeZiminski:
This PR does a basic check on the `top_level_folder` input to `create_folders` and `upload_custom` / `download_custom`. These three methods it is easy to accidently put the subject / session in first.

This is not tested as very small, and is essentially a sticking plaster for a better way of doing argument validation outlined in #380. No documentation is required.

### DIFF
--- a/datashuttle/datashuttle.py
+++ b/datashuttle/datashuttle.py
@@ -1459,7 +1459,7 @@ class DataShuttle:
 
     def _check_top_level_folder(self, top_level_folder):
         """
-Raise an error if ``top_level_folder`` not correct.
+        Raise an error if ``top_level_folder`` not correct.
         """
         canonical_top_level_folders = canonical_folders.get_top_level_folders()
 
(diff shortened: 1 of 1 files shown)


datashuttle/datashuttle.py
```diff
@@ -1451,3 +1456,16 @@ def _update_settings_with_new_canonical_keys(self, settings: Dict):
         for key in ["overwrite_existing_files", "dry_run"]:
             if key not in settings["tui"]:
                 settings["tui"][key] = canonical_tui_configs["tui"][key]
+
+    def _check_top_level_folder(self, top_level_folder):
+        """
+        Raise an error if `top_level_folder` not correct.
```
niksirbi, line 1462: Needs doube backticks to rended as monospace in sphinx. As it is now, the result will be italicised ( I think). Not that it matters much, since this is a private functions.

```suggestion
Raise an error if ``top_level_folder`` not correct.
````

niksirbi (approved):

PR #382, opened 2024-04-29
Create folders return split by datatype

### LINKED ISSUE #374: `create_folders` should split returned paths by datatype
Currently `create_folders` returns a big list of paths where datatypes are mixed.

It would be much better to split these by datatype, returning a dict of lists of paths `{"behav": [], "ephys": []}

[This function](https://github.com/neuroinformatics-unit/datashuttle/blob/aea2bb99dbe4a789e476e64b10cc9eae364a4da3/datashuttle/utils/folders.py#L67) needs to be changed.

JoeZiminski:
This PR splits the output of `create_folders` by datatype. Now a `Dict` is returned with datatypes as keys and a list of created paths as values. If no datatype is created the dict keys are `sub` or `ses`.

Test and docs are updated. closes  #374.


tests/test_utils.py
```diff
@@ -235,6 +235,9 @@ def check_folder_tree_is_correct(
     rely on project settings itself,
     as this doesn't explicitly test this.
     """
+    if created_folder_list is None:
```
niksirbi, line 238: shouldn't this be renamed to `created_folder_dict`?

datashuttle/datashuttle.py
```diff
@@ -148,7 +148,7 @@ def create_folders(
         ses_names: Optional[Union[str, List[str]]] = None,
         datatype: Union[str, List[str]] = "",
         bypass_validation: bool = False,
-    ) -> List[Path]:
+    ) -> Dict[str, Path]:
```
niksirbi, line 151: this function is missing a "Returns" docstring section. Perhaps you can add it, describing the contents of the returned dict?

docs/source/pages/tutorials/getting_started.md
```diff
@@ -450,7 +450,7 @@ folder_path_list = project.create_folders(
 
 )
 
-print([path_ for path_ in folder_path_list if path_.name == "behav"])
+print([path_ for path_ in folder_path_list["behav"]])
```
niksirbi, line 453: rename `folder_path_list` to `folder_path_dict` within this code block

datashuttle/utils/folders.py
```diff
@@ -89,12 +94,16 @@ def create_folder_trees(
...
+                    datatype,
+                    ses_path,
+                    "ses",
+                    save_paths=all_paths,
+                    log=log,
                 )
-                all_paths.extend(datatype_paths)
             else:
-                all_paths.append(ses_path)
+                all_paths["ses"].append(sub_path)
```
niksirbi, line 106: Quite sure you intended to use `ses_path` here.
```suggestion
                all_paths["ses"].append(ses_path)
```

tests/tests_integration/test_create_folders.py
```diff
@@ -252,6 +252,9 @@ def test_datetime_flag_in_session(self, project):
         assert all([re.search(datetime_regexp, name) for name in ses_names])
         assert all([tags("time") not in name for name in ses_names])
 
+    def test_error_raised_with_skipped_session_input(self, project):
+        pass
```
niksirbi, line 256: why create a test that does nothing? did you intende to fill it up later?

datashuttle/utils/folders.py
niksirbi: I guess this wasn't caught because this particular case (splitting output by sub/ses) is not tested (unless I missed it).

niksirbi (changes requested):
Hey @JoeZiminski, I spotted a few issues that slipped through in this one. Luckily they are trivial to fix.

tests/tests_integration/test_create_folders.py
JoeZiminski: good spot! In the end I don't think that test is worth doing. It might be addressed soon anyway if we allow skipping the session level.

JoeZiminski:
Thanks @niksirbi for these catches 😱 a lot of mistakes! These should all be fixed now, I added a test to cover the previously broken behaviour. Cheers!

niksirbi (approved):
No worries, that's why re review all PRs 😉 
LGTM now.

PR #384, opened 2024-05-01
Reworking logging.

JoeZiminski:
This PR rewords how the logging works due to a regression in which the root logger rather than a named `"datashuttle"` logger was used. I am quite confused as to what happened here as `fancylog` has never created a named logger. It happened to work because `"datashuttle"` logger was propagating through the `FileHandler` stream on the root logger that `fancylog` had setup. 

Anyway, this PR:

1) Is based on recent fancylog changes (v0.4.2) which allow starting a named logger (i.e. `logger=getLogger("some_name")` rather than `logger=getLogger()` which gets the root).
2) Now all interaction with the logger in datashuttle is ensured to be through this named logger. This is supported by a number of convenience functions in `ds_logger.py`.
3) Add tests that the logger is setup and logging as expected.

This PR requires #387 for tests to pass, as new version of `fancylog` requires >=Python3.9. It requires no further tests or documentation changes.

### DIFF
--- a/datashuttle/utils/ds_logger.py
+++ b/datashuttle/utils/ds_logger.py
@@ -27,9 +27,8 @@ def get_logger():
 
 def logging_is_active():
     logger_exists = get_logger_name() in logging.root.manager.loggerDict
-    if logger_exists:
-        if get_logger().handlers != []:
-            return True
+    if logger_exists and get_logger().handlers != []:
+        return True
     return False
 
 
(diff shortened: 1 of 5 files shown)


niksirbi (approved):
That sounds scary, I don't think I fully understand loggin in Python yet.
That aside, this looks good to me. I played around creating some projects and the produced logs were as expected.

JoeZiminski:
Yes @niksirbi I was quite surprised as to how it happened, I am still finding new ways for the `logging` module to confuse me.  I think in this case the the take home message was to always use a named logger!

PR #400, opened 2024-06-13
Remove unused sphinx-argparse dependency.

JoeZiminski:
This PR removes the now unused `sphinx-argparse` dependency (that was left over from when we had CLI / CLI docs).

Tested by building locally, everything worked without any new warnings.


niksirbi (approved):
LGTM 🎉

PR #414, opened 2024-09-24
Make CI run with weekly cron job.

JoeZiminski:
This PR updates the CI test framework to also run on a cron job once a week. A nice benefit of this is we can cut down the python versions tested for typical CI runs (now only oldest + most recent python) but also have the weekly check across all versions to asuage any paranoia! #bestofbothworlds.


adamltyson:
Naive question - why do we need an issue opened when a test fails? An email will be sent anyway. Issues may just spam any watchers of the repo.

JoeZiminski:
No it's a good question! I didn't expect that to happen. I psuhed something that should now only create one issue if any test fails. This is on a cron job to run on main once a week, and I'd hope that it will fail very rarely. The benefit of an issue is it can't be lost in inboxes. However also good point, I think ATM it will spawn cron jobs from all branches / forks (?). Will look into only running from main, and not from forks

adamltyson:
I think the problem with issues is that they aren't informative, it just says "failed". These auto generated issues will need to be replaced with a more informative issue anyway.

JoeZiminski:
Hmm yes, this is also super annoying that you can't delete issues so they are just there spamming up the closed issues forever. Maybe email / github notification is sufficient, it it turns out these are being mised often, then we can increase cron job frequency / look into spawning issues.

JoeZiminski:
Well, actually the organisation owner can delete issues, so I will probably annoy you more by asking you to do this when your back 😆 but I've removed the issue-making at least so there should be no more.

JoeZiminski:
I'll merge for now to make a release for the course but happy for feedback anytime on this

PR #446, opened 2024-12-04
Fix tests after textual changes

JoeZiminski:
The (private) `_text` attribute I was using to check label contents in tests was changed in recent textual version. This is now updated from `renderable._text[0]` to simply `renderable`.

Tests are failing, this is due to out-of-bound error issue in test suite only. For now will merge to incorporate these usueful fixes and follow up with [this issue](https://github.com/Textualize/textual/issues/4412).


niksirbi (approved):
Good to go!

PR #457, opened 2025-02-04
Add local only mode

JoeZiminski:
This PR loosens the requirements to pass a full set of configs that specify, in addition to the path where the local project resides, the path to the central project and method of connection.

This makes it much easier to get started with `datashuttle`, it is enough to pass a single path to get access to project validation and folder creation features.

In this new 'local only' mode, some functionality is not available (e.g. data transfer, and and the `local_only=False` functionality of certain methods such as `get_next_sub` or `validate_project`. For example, now `get_next_sub` will only search the local project even if `local_only=False`, because the central project details have not been set up. I hope this is fairly intuitive, even though it means there is an argument on some functions when in 'local_mode' that doesn't do anything. I don't think anyone who had not put details of a central project would expect it to be scraped for details (?).

To introduce this new mode, this PR:
- allows `central_path` and `connection_method` to be `None`. Both must be `None` or both must be set. I also refactored some very messy config functions as a part of this, which was slightly tangential :(
- Introduces a decorator that throws an error if the project is 'local_only' (this is used to make sure access to functionality e.g. transfer which requires a central project is blocked
- Updates the TUI to handle this case. There is now a new radiobutton for connection method - 'No connection (local only)'. When this is set, central path input widget is greyed out. It is possible to add `central_path` and `connection_method` to the project . If the project is `local_mode`,  the transfer tab is greyed out. When switching between local and full project modes (by adding / removing the required configs), the ProjectManager screen must be refreshed.
- Add tests covering python api and tui ends. For testing the local-mode project, the fixture that is used in tests was adjusted to return either a local-only or full project, depending of a indirect `pytest.mark.parameterize`. This means with this decorator we can test the local-only case for free. I added this in a lot of places, maybe it is overkill, but it doesn't add much to testing time. 
- Add documentation.  I think the documentation is slightly confusing, the whole idea of a 'local only project' and a 'full project' should probably not be exposed to the user. But the docs are due a major rework soon to be simpler / more accessible so planning to address in more detail there.

### Testing Manually

For code, this can be tested like:

```
from datashuttle import DataShuttle

project = DataShuttle("my_project")

project.make_config_file(local_path="/path/to/project/")

# then you can do stuff like

project.create_folders("rawdata", "sub-001", ...)
project.validate_project()

# should see an error for 
project.upload_entire_project()
```
[API reference](https://datashuttle.neuroinformatics.dev/pages/api_index.html)

For the TUI, you can do:

```
datashuttle launch
```

and then 'Create New Project' and just enter the local path with the 'No connection'  radiobutton (then you can go to project and create some folders, validation is not exposed here yet).


datashuttle/configs/load_configs.py
```diff
@@ -59,3 +60,37 @@ def attempt_load_configs(
         )
 
     return new_cfg
+
+
+def convert_str_and_pathlib_paths(
```
JoeZiminski, line 65: this was a straight refactor, unrelated to the aims of this PR :(

datashuttle/configs/config_class.py
```diff
@@ -121,44 +122,14 @@ def load_from_file(self) -> None:
...
-        self.convert_str_and_pathlib_paths(config_dict, "str_to_path")
+        load_configs.convert_str_and_pathlib_paths(config_dict, "str_to_path")
 
         self.data = config_dict
 
     # -------------------------------------------------------------------------
     # Utils
     # -------------------------------------------------------------------------
 
-    def convert_str_and_pathlib_paths(
```
JoeZiminski, line 132: This is a refactor, moving to another module to avoid the strange signature that was occuring:

```configs.convert_str_and_pathlib_paths(configs, ...)```

File: tests/tests_tui/test_local_only_project.py [hunk 3]
```diff
@@ -0,0 +1,233 @@
...
+            )
+            assert pilot.app.screen.interface.project.cfg == {
+                "local_path": local_path / project_name,
+                "central_path": None,
+                "connection_method": None,
+                "central_host_id": None,
+                "central_host_username": None,
+            }
+            assert pilot.app.screen.query_one(
+                "#placeholder_tranfser_tab"
```
IgorTatarnikov, line 42: Is this a typo? "#placeholder_transfer_tab" instead?

IgorTatarnikov (approved):
Just a few tiny comments but it looks great overall.

I was able to use the TUI and the python API to create and work with a local only project.

PR #458, opened 2025-02-05
Fix copy keyboard shortcut

JoeZiminski:


adamltyson:
Not sure about this. I expect Ctrl+C to quit, but maybe many users will expect it to mean copy.

JoeZiminski:
Yes @adamltyson I can't decide either. I was actually learning towards `CTRL+C` for copy, as the TUI is for people less familiar with programming anyway. However, I can still force the application to quit with CTRL+C when it should not be possible (under discussion [here](https://github.com/Textualize/textual/issues/5503)).

As this is resulting in failing tests, and this PR only maintains current behaviour, do you think it's okay to merge now? and we can revisit the best button to use when the discussion on the above issue is finished.

adamltyson:
Yeah go for it. The problem can be fixed to some extent with good docs.

PR #460, opened 2025-02-10
Add additional message when quitting SSH setup.

JoeZiminski:
Just an additional message to confirm SSH setup was quit. I got confused at this stage and was entering my password here rather than pressing 'y' and SSH was silently not being set up.

NOTE: tests are failing due to known issue in #458 so I think okay to merge.


adamltyson (approved):
Approving, but not merging due to test failures.

Issue #467, opened 2025-02-28
GUI crashes after naming subject (manually or with double-click)


Malkemperlab:
I am running datashuttle the first time on a Windows PC, using the Anaconda prompt. After creating a new project, I would like to run a test, but the GUI always quits after entering a subject or session name and gives the following error. Help would be much appreciated. 

╭────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── Traceback (most recent call last) ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
│ C:\Users\Malkemper\AppData\Local\anaconda3\envs\datashuttle-env\Lib\site-packages\datashuttle\tui\utils\tui_decorators.py:33 in wrapper
│
│   30 │   │                                                                                    ╭──────────────────────────────── locals ─────────────────────────────────╮
│   31 │   │   if click_time - parent_class.prev_click_time < 0.5:                              │         args = (CreateFoldersTab(id='tabscreen_create_tab'), Clicked()) │
│   32 │   │   │   parent_class.prev_click_time = click_time                                    │   click_time = 27202.406                                                │
│ ❱ 33 │   │   │   return func(*args, **kwargs)                                                 │       kwargs = {}                                                       │
│   34 │   │                                                                                    │ parent_class = CreateFoldersTab(id='tabscreen_create_tab')              │
│   35 │   │   parent_class.prev_click_time = click_time                                        ╰─────────────────────────────────────────────────────────────────────────╯
│   36
│
│ C:\Users\Malkemper\AppData\Local\anaconda3\envs\datashuttle-env\Lib\site-packages\datashuttle\tui\tabs\create_folders.py:138 in on_clickable_input_clicked
│
│   135 │   │   if event.ctrl:                                                                   ╭──────────────────────── locals ────────────────────────╮
│   136 │   │   │   self.fill_input_with_template(prefix, input_id)                              │    event = Clicked()                                   │
│   137 │   │   else:                                                                            │ input_id = 'create_folders_session_input'              │
│ ❱ 138 │   │   │   self.fill_input_with_next_sub_or_ses_template(prefix, input_id)              │   prefix = 'ses'                                       │
│   139 │                                                                                        │     self = CreateFoldersTab(id='tabscreen_create_tab') │
│   140 │   def on_custom_directory_tree_directory_tree_special_key_press(                       ╰────────────────────────────────────────────────────────╯
[...]

PR #468, opened 2025-02-28
Pin textual version

JoeZiminski:
#467 is due to a textual update. Textual is still updating [very quickly](https://pypi.org/project/textual/#history), I'd hoped to avoid pinning a version but I think this is the only sustainable solution for now. We can periodically review and test new textual versions.


adamltyson:
Seems sensible to me. If it's just the one, key package it shouldn't be too hard to periodically keep it up to date.

adamltyson (approved):

Issue #466, opened 2025-02-21
Validate broad vs. narrow datatypes


JoeZiminski:
In [NeuroBlueprint](https://neuroblueprint.neuroinformatics.dev/latest/specification.html#datatype) if you have one narrow datatype (e.g. `icephys` instead of `ephys`) then you cannot use the broad datatype at all in the project, and must use the narrow datatypes for that datatype. Currently this is not validated in `datashuttle`.

aakash-test7:
Query :
(1) Priority to be given to first data-type if its narrow then restrict to use of narrow data-type only, if possible convert broad data-type to narrow if not then display error message and urge for change by user end .
(2) Both data-types to be checked if first one is broad and second is narrow then too apply the same logic to convert to narrow and so.

Thankyou in advance 
I'll try to get this done . Looking the code

JoeZiminski:
It might make sense to combine fixing this function with #516.

In general this validation is quite awkward. It will require collecting all valid datatype folders across the entire project (local and central, if central is included). Then it will require one pass over this. This will be very slow for central connection, as will involve checking each session folder. I think just add this to `strict_mode` which doesn't include central.

Issue #449, opened 2024-12-17
Check that `central_path` can be written to during SSH setup.


JoeZiminski:
It is very easy to get the wrong path on the central server as it is often not obvious what the full path is. For example, on our hpc the full path is like `/ceph/nfs/home/jzimiminski` etc. and it is not always clear what the path should be. During 'setup ssh connection' it would be nice to have a new screen to a) confirm the full path of the central path relative to the server b) check you can actually write there. This has come up in user testing and is a priority.

aakash-test7:
After going through your suggestions, I have 2 approaches. In one, move the "check_write_permission" to utils/ssh.py and then assume user has input path correct. this remove the state of freeze. In another approach, I will try to merge "check_write_permission" with "_setup_rclone_central_ssh_config" . this may even remove the central path condition from setup_ssh_connection and move It to "_setup_rclone_central_ssh_config". Rest let's see how it goes.

JoeZiminski:
Given the extension to AWS and Google Drive, the new-favourite approach is to use rclone to check permissions as it will work across all methods. It could be done after set up and raise a warning / error if permissions are not found. The approach could be:

```
rclone touch remote:path/.rclone_write_test
```
then
```
rclone delete remote:path/.rclone_write_test
```

JoeZiminski:
This is now complete with #503 and introduction of the `check_successful_connection_and_raise_error_on_fail` function

PR #464, opened 2025-02-18
Extend validation

JoeZiminski:
This PR extends and refactors the validation (`validation.py`):
- error messages are better standardized, each containing some key that describes the error (e.g. `DUPLICATE_NAME`) as well as a path to the folder causing the issue.
- `validation.py` is completely refactored, its still a little complex because of some annoying edge-cases but the logical all flows start-to-finish rather with recursion removed.
- public validation functions now return `error_messages` which contains a list of strings of all validation errors.
- A couple more validation cases caught.
- a `strict_mode` argument added. This will check all folders in the project and if they are not in a datatype folder ensures they are NeuroBlueprint. Currently, the validate will search for sub- and ses- prefixed names and check these only. Datatypes will not be validated. The problem is that NeuroBlueprint allows folders that are not NeuroBlueprint formatting for flexibility, so for example with datatype, we do not know if a folder is a bad datatype folder or just some other folder the user wants. So its not possible to validate. But in strict mode, we can validate everything. 
- docs and tests added

To test, this can be run with a file path `quick_validate_project` or to also get logs with `project.validate_project(...)`:
```
path_to_project = "path/to/project"
project_name = "some_name"

from datashuttle import quick_validate_project
from datashuttle import DataShuttle

# quick way
error_messages = quick_validate_project(path_to_project)

# datashuttle-project way (will output logs)

project = DataShuttle(project_name)
# first time: project.make_config_file(local_path=path_to_project)

error_messages = project.validate_project(..., strict_mode=...)

```

To check the docs (after `pip install -r docs/requirements.txt`)

```
`sphinx-build source build
```

and there is a new 'validation` page on the tutorials.

Unfortunately because this contains a lot of refactoring, the diff is quite big. The main changes are the refactoring in `validation.py` and some associated tests. It might be easiest just to look at the new `validation.py` file, unfortunately the diff is quite long sorry.

NOTE: tests are failing on linux / macos (just trying windows) but I cannot replicate locally on windows. Will check on macOS asap.

PR #469, opened 2025-03-01
Aakash/449

aakash-test7:
Enhancement for issue #466 
Enhancement for issue #449

### DIFF
--- a/datashuttle/datashuttle_class.py
+++ b/datashuttle/datashuttle_class.py
@@ -868,10 +868,31 @@ class DataShuttle:
 
         if verified:
             ssh.setup_ssh_key(self.cfg, log=True)
+            #449
+            confirmed_path = input(
+                f"Confirm the full central path (default: {self.cfg['central_path']}): "
+            ) or self.cfg["central_path"]
+            self.cfg["central_path"] = confirmed_path
+            if not self._check_write_permissions(confirmed_path):
+                raise PermissionError(f"Cannot write to the central path: {confirmed_path}")
+            
             self._setup_rclone_central_ssh_config(log=True)
 
         ds_logger.close_log_filehandler()
 
+    def _check_write_permissions(self, path) -> bool:
+        """
+        Check if the user has write permissions on the central path.
+        """
+        try:
+            # Use SSH to check write permissions on the remote server
+            command = f"touch {path}/test_write_permission.txt && rm {path}/test_write_permission.txt"
+            result = ssh.run_ssh_command(self.cfg, command, log=False)
+            return result.returncode == 0
+        except Exception as e:
+            print(f"Write permission check failed: {e}")
+            return False
+        
     @requires_ssh_configs
     @check_is_not_local_project
     def write_public_key(self, filepath: str) -> None:
--- a/datashuttle/datashuttle_class.py
+++ b/datashuttle/datashuttle_class.py
@@ -885,12 +885,12 @@ class DataShuttle:
 
         ds_logger.close_log_filehandler()
 
+    # 449
     def _check_write_permissions(self, path) -> bool:
         """
         Check if the user has write permissions on the central path.
         """
         try:
-            # Use SSH to check write permissions on the remote server
             command = f"touch {path}/test_write_permission.txt && rm {path}/test_write_permission.txt"
             result = ssh.run_ssh_command(self.cfg, command, log=False)
             return result.returncode == 0
--- a/datashuttle/datashuttle_class.py
+++ b/datashuttle/datashuttle_class.py
@@ -868,14 +868,19 @@ class DataShuttle:
 
         if verified:
             ssh.setup_ssh_key(self.cfg, log=True)
-            #449
-            confirmed_path = input(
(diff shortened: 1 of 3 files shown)


parharti:
Hey! It looks like the pre-commit checks might be failing. When you get a chance, you could try running pre-commit run --all-files and then commit again Thanks a lot!

aakash-test7 (commented):
issues #466 and #449

aakash-test7:
Thanks a lot @parharti.

JoeZiminski:
Thanks for this @aakash-test7 this is much appreciated. 

For the changes related to #466, would you be able to split this off into a separate PR? It would make sense to wait until #464 is merged and then we can focus on this.

In this PR I will review the changes related to #469, thanks again! and cheers @parharti for the help with pre-commit

datashuttle/datashuttle_class.py
```diff
@@ -868,10 +868,36 @@ def setup_ssh_connection(self) -> None:
 
         if verified:
             ssh.setup_ssh_key(self.cfg, log=True)
+            # 449
+            confirmed_path = (
```
JoeZiminski, line 872: This is a nice idea, but taking an input here would freeze the application when run through the TUI. At this stage we can trust the user has input the path correctly at the config stage (if they have not, they will realize after a permission error below). Here I would delete the `input` code and use `self.cfg["central_path"]` directly in `self._check_write_permissions`

datashuttle/datashuttle_class.py
```diff
@@ -868,10 +868,36 @@ def setup_ssh_connection(self) -> None:
...
+                raise PermissionError(
+                    f"Cannot write to the central path: {confirmed_path}"
+                )
+
             self._setup_rclone_central_ssh_config(log=True)
 
         ds_logger.close_log_filehandler()
 
+    # 449
+    def _check_write_permissions(self, path) -> bool:
```
JoeZiminski, line 889: This is cool, I think this function can go into `ssh.py` and be merged with `run_ssh_command`. It is nice to encapsulate the code for writing commands over SSH but as we do this [elsewhere](https://github.com/neuroinformatics-unit/datashuttle/blob/166f8620199b6392df43d0889a8b5e34b263d3a0/datashuttle/utils/ssh.py#L66) in a way that would make it difficult to use `run_ssh_command`.  For now we can use `client.exec_command` more flexibly from within other functions and refactor to centralise under a single function at a later time.

datashuttle/datashuttle_class.py
```diff
@@ -868,10 +868,36 @@ def setup_ssh_connection(self) -> None:
...
             ssh.setup_ssh_key(self.cfg, log=True)
+            # 449
+            confirmed_path = (
+                input(
+                    f"Confirm the full central path (default: {self.cfg['central_path']}): "
+                )
+                or self.cfg["central_path"]
+            )
+            self.cfg["central_path"] = confirmed_path
+            if not self._check_write_permissions(confirmed_path):
```
JoeZiminski, line 879: This could be moved under `_setup_rclone_central_ssh_config` so that the connection is set up fully, but then the user is informed if the central path they have chosen does not have the permission error (it is possible to set up the connection even if the central path is wrong). So the order can be like

1) setup the connection fully and message to indicate the connection is okay.
2) raise if the central path does not have permissions

datashuttle/utils/ssh.py
```diff
@@ -320,3 +321,47 @@ def get_list_of_folder_names_over_sftp(
...
+) -> SSHCommandResult:
+    """
+    Run a command on the remote server over SSH and return a result object.
+    """
+    client = paramiko.SSHClient()
+    try:
+        connect_client_with_logging(
+            client, cfg, password, message_on_sucessful_connection=log
+        )
+        stdin, stdout, stderr = client.exec_command(command)
```
JoeZiminski, line 348: This is nice, in this case I think we can test directly the permissions with `sftp.open()` and only log an error related to failed permissions in the case the error is raised. Combining with the function above (`_check_write_permissions`), what do you think of a function like this?

in `ssh.py`:

```
def raise_on_bad_write_permissions(
        cfg: Configs,
) -> None:
    with paramiko.SSHClient() as client:

        connect_client_with_logging(client, cfg, message_on_sucessful_connection=False)

        try:
            with sftp.open(cfg["central_path"] / "test_write_permission", "w") as file:
                file.write("test")
            sftp.remove(test_file)
        except IOError:
            utils.log_and_raise_error(
                f"Could not write files in the central project path. Please check permissions at: {cfg['central_path']}",
                ConnectionError,
            )

```

I removed the logging option here, as I realized all of the `ssh` functions expose a logging parameter, but it is only ever set as `True`. It adds quite a lot of complexity to juggle logging on / off, and if it is always on, we might as well remove the option of `log=False` across the `ssh` module in a follow up PR! Thanks

JoeZiminski (changes requested):
Thanks @aakash-test7! your additions were a great solution to the problem. I have suggested some refactorings, let me know what you think.

Issue #459, opened 2025-02-07
Attempting to copy in the terminal user interface crashes on headless HPC


JoeZiminski:
Most HPC systems are run in 'headless' move, without X11 windowing system. This means that certain functionality that X11 exposes e.g. clipboard is not available. When attempting to copy, the below is raised:
```
PyperclipException:
    Pyperclip could not find a copy/paste mechanism for your system.
    For more information, please visit https://pyperclip.readthedocs.io/en/latest/index.html#not-implemented-error
```

All suggestions on the link above do not work in headless mode because there is no X11 (e.g. so `xclip` doesn't work). X11 forwarding I think should be a workaround, or using `xvfb` (but requires sudo). Basically none of this can be managed on the `datashuttle` side, so we should catch the `PyperclipException` and show a dialog explaining the problem.

Issue #442, opened 2024-11-28
Add `get_subs` `get_ses` and other useful getters


JoeZiminski:
datashuttle should be an easy way to run your experiment, not just collect data. There are currently not many convenience functions like getter / setter for paths. It would be really nice to do something like:

```
project_info = project.get_all_sub_and_ses()

for sub_info in project_info:
    for ses_path in sub_info["session_paths"]:
          
          ...
```

or like 

```
shank_paths = project.find_folders("sub-001", "all_ses")
....
```

This wouldn't be much to implement but would expand the utility a lot, at the moment it's not really clear how to use datashuttle as a worflow manager.

PR #470, opened 2025-03-01
Attempting to copy in the terminal user interface crashes on headless HPC #459

parharti:
In this PR, I have implemented a dialog box that appears when a clipboard copy operation fails on headless computers. Instead of printing an error message, the application will now display a user-friendly dialog informing the user about the issue. This improves usability by ensuring that users are aware of clipboard limitations in headless environments as suggested in the issue.


datashuttle/tui/custom_widgets.py
```diff
@@ -39,6 +42,19 @@
 # --------------------------------------------------------------------------------------
 
 
+class ClipboardErrorDialog(ModalDialog):
```
JoeZiminski, line 45: This is a nice solution! We have something very similar in `modal_dialogs.MessageBox` and a mainwindow function `show_modal_error_dialog` and so it makes sense to use these if possible, it would be something like:

```
            try:
                pyperclip.copy(self.value)
            except pyperclip.PyperclipException:
                self.mainwindow.show_modal_error_dialog(
                    "Clipboard copy failed, likely due to operating in headless mode (e.g. on a high-performance computer)."
               )
  ```

File: tests/tests_tui/test_tui_directorytree.py [hunk 2]
```diff
@@ -131,7 +131,10 @@ async def test_create_folders_directorytree_clipboard(
             await self.reload_tree_nodes(
                 pilot, "#create_folders_directorytree", 4
             )
-            pyperclip.copy("STARTING VAL")
+            try:
```
JoeZiminski, line 134: Here I think in the test environment, we should never be in headless mode (tests are running okay in the CI). In this case, we can let a `PyperclipException` be raised just so we are aware of any other possible failures.

tests/tests_tui/test_tui_directorytree.py
```diff
@@ -131,7 +131,10 @@ async def test_create_folders_directorytree_clipboard(
...
+                pyperclip.copy("STARTING VAL")
```
JoeZiminski, line 135: Here I think in the test environment, we should never be in headless mode (tests are running okay in the CI). In this case, we can let a `PyperclipException` be raised just so we are aware of any other possible failures.

datashuttle/tui/custom_widgets.py
JoeZiminski, line 45: This is a nice solution! We have something very similar in `modal_dialogs.MessageBox` and a mainwindow function `show_modal_error_dialog` and so it makes sense to use these if possible, it would be something like:

```
            try:
                pyperclip.copy(self.value)
            except pyperclip.PyperclipException:
                self.mainwindow.show_modal_error_dialog(
                    "Clipboard copy failed, likely due to operating in headless mode (e.g. on a high-performance computer)."
               )
  ```

**Review by JoeZiminski — CHANGES_REQUESTED** [review]
Thanks a lot for this @parharti, it looks great. Just a couple of comments related to using some existing functionality. Cheers!

File: tests/tests_tui/test_tui_directorytree.py (hunk 3 above)
Comment (parharti) [reply tests/tests_tui/test_tui_directorytree.py]: Got it! If there is an unexpected issue with clipboard operations, the test will fail, making debugging easier and ensuring we catch any other possible failures.

**Comment by JoeZiminski** [comment]
Thanks @parharti, this is looking great! There is one other use of `pyperclip` [here](https://github.com/neuroinformatics-unit/datashuttle/blob/264300efff5608dc1b40dbe1ab0aa92532f9a9cb/datashuttle/tui/custom_widgets.py#L127). We could wrap this in another try/except and copy the message but it might lead to some duplication. Because both of these classes hold the `mainwindow`, I think it makes to move the copying to `mainwindow`. So we could have a function on `app.py` something like:

```
def copy_to_clipboard():
    try:
        ....
```

and then from within the classes, do `self.mainwindow.copy_to_clipboard(<value to copy>)`. What do you think?

I was thinking about how to test this, as its quite tricky to test as it only appears under certain circumstances.

However, I think we can use monkeypatching to check this case. We can mock the `pyperclip.copy` function such that it always returns `PyperclipException` and then check this is caught in the TUI.

We can add the test under this test [here](https://github.com/neuroinformatics-unit/datashuttle/blob/f12971366f917d5fa634f0909c9fa681722bac09/tests/tests_tui/test_tui_directorytree.py#L114).

A test for this would look like:

```
@pytest.mark.asyncio
    async def test_failed_pyperclip_copy(
            self, setup_project_paths, monkeypatch
    ):
        tmp_config_path, tmp_path, project_name = setup_project_paths.values()

        app = TuiApp()
        async with app.run_test(size=self.tui_size()) as pilot:

            # Set up a project and navigate to the directory tree
            await self.setup_existing_project_create_tab_filled_sub_and_ses(
                pilot, project_name, create_folders=True
            )

            await self.reload_tree_nodes(
                pilot, "#create_folders_directorytree", 4
            )

            # Monkeypatch the pyperclip copy function
            def mock_copy(_):
                raise PyperclipException("Mock function error raised.")

            monkeypatch.setattr(pyperclip, "copy", mock_copy)

            # Trigger a copy on the TUI such that the mocked copy error is raised
            await self.hover_and_press_tree(
                pilot,
                "#create_folders_directorytree",
                hover_line=2,
                press_string="ctrl+q",
            )

            # Check that the error message is displayed.
            assert "Clipboard copy failed, likely due to operating in headless mode" in pilot.app.screen.query_one(
                    "#messagebox_message_label"
                ).renderable
```

**Review by JoeZiminski — CHANGES_REQUESTED** [review]
Thanks @parharti, this is looking great! I noticed there is one other use of `pyperclip` [here](https://github.com/neuroinformatics-unit/datashuttle/blob/264300efff5608dc1b40dbe1ab0aa92532f9a9cb/datashuttle/tui/custom_widgets.py#L127). We could wrap this in another try/except and copy the message but it might lead to some duplication. Because both of these classes hold the `mainwindow`, I think it makes to move the copying to `mainwindow`. So we could have a function on `app.py` something like:

```
def copy_to_clipboard():
    try:
        ....
```

and then from within the classes, do `self.mainwindow.copy_to_clipboard(<value to copy>)`. What do you think?

I was thinking about how to test this, as its quite tricky to test as it only appears under certain circumstances.

However, I think we can use monkeypatching to check this case. We can mock the `pyperclip.copy` function such that it always returns `PyperclipException` and then check this is caught in the TUI.

We can add the test under this test [here](https://github.com/neuroinformatics-unit/datashuttle/blob/f12971366f917d5fa634f0909c9fa681722bac09/tests/tests_tui/test_tui_directorytree.py#L114).

A test for this would look like:

```
@pytest.mark.asyncio
    async def test_failed_pyperclip_copy(
            self, setup_project_paths, monkeypatch
    ):
        tmp_config_path, tmp_path, project_name = setup_project_paths.values()

        app = TuiApp()
        async with app.run_test(size=self.tui_size()) as pilot:

            # Set up a project and navigate to the directory tree
            await self.setup_existing_project_create_tab_filled_sub_and_ses(
                pilot, project_name, create_folders=True
            )

            await self.reload_tree_nodes(
                pilot, "#create_folders_directorytree", 4
            )

            # Monkeypatch the pyperclip copy function
            def mock_copy(_):
                raise PyperclipException("Mock function error raised.")

            monkeypatch.setattr(pyperclip, "copy", mock_copy)

            # Trigger a copy on the TUI such that the mocked copy error is raised
            await self.hover_and_press_tree(
                pilot,
                "#create_folders_directorytree",
                hover_line=2,
                press_string="ctrl+q",
            )

            # Check that the error message is displayed.
            assert "Clipboard copy failed, likely due to operating in headless mode" in pilot.app.screen.query_one(
                    "#messagebox_message_label"
                ).renderable
```

There is quite a lot going on in this test so please let me know if you have any questions.

JoeZiminski (changes requested):
Hi @parharti thanks for this! it is nearly ready to merge I think, please see a few comments above.

This PR has a couple of commits on it from PR #442, I am having some trouble accessing this branch on your fork. I think this is because it is called `Prisha` and the branch for #442 is called `prisha` and git is not distinguishing these two branches. In general the branch names can be chosen to reflect the changes in the branch in a concise way (e.g. `fix_copy_on_headless`).

There are two ways to solve this, if locally you can access your `Prisha` branch with the commits for this PR, when you are on it, you can `git checkout -b fix_copy_on_headless` and then open a new PR and close this one (you can just link to the old one in the new one, it's no problem this kind of thing happens quite often). You can use `git reset` to remove the few extra commits at the end of this PR which are unrelated.

Otherwise, if the `prisha` branch has overwritten `Prisha` for you locally, it is possible to make a new branch and cherry-pick the relevant commits, though this will be a little more work as there are quite a few commits to cherry pick.

parharti:
> Hi @parharti thanks for this! it is nearly ready to merge I think, please see a few comments above.
> 
> This PR has a couple of commits on it from PR #442, I am having some trouble accessing this branch on your fork. I think this is because it is called `Prisha` and the branch for #442 is called `prisha` and git is not distinguishing these two branches. In general the branch names can be chosen to reflect the changes in the branch in a concise way (e.g. `fix_copy_on_headless`).
> 
> There are two ways to solve this, if locally you can access your `Prisha` branch with the commits for this PR, when you are on it, you can `git checkout -b fix_copy_on_headless` and then open a new PR and close this one (you can just link to the old one in the new one, it's no problem this kind of thing happens quite often). You can use `git reset` to remove the few extra commits at the end of this PR which are unrelated.
> 
> Otherwise, if the `prisha` branch has overwritten `Prisha` for you locally, it is possible to make a new branch and cherry-pick the relevant commits, though this will be a little more work as there are quite a few commits to cherry pick.

To resolve the conflict, I deleted the prisha branch and will create a new one. Right now, my Prisha branch is up to date with origin/Prisha, so I'll check if everything works as expected. If any issues remain, I'll open a new branch and PR. I'll try my best to save this branch, as working with Git is always a learning experience!

JoeZiminski (approved):
Hey @parharti thanks for these changes, LGTM! The tests are all working, and I checked manually (normal copy / paste works well and the error shows nicely in headless mode on the HPC). Yes `git` has a lot of 'gotchas' and it takes a while to find them all out! After your changes the  branch worked well for me locally.

Cheers for this, will run the tests now and merge tomorrow!

PR #472, opened 2025-03-02
[Fix #438] Disallow empty project names

### LINKED ISSUE #438: Disallow empty file name in TUI project setup
At the moment (surprisingly) it is possible to create a project with an empty name. Rare this would happen in practice but should definitely not be allowed, maake a conditional for this case and throw error.

chandrikarj:
## Description
Fixes issue #438 : Disallow empty file name in TUI project setup

## What is this PR

- [x] Bug fix
- [ ] Other

## Why is this PR needed?

Currently, the TUI allows users to create a project with an empty name, which can lead to unexpected behavior. 
This PR introduces a validation step to ensure that project names cannot be empty. If an empty name is detected, an error message is displayed, preventing project creation.

## What does this PR do?

1. Added a check in the project creation logic to verify if the name is empty.
2. Updated the UI to display a clear error message when this occurs.

_Demonstration of the changes:_
https://github.com/user-attachments/assets/6b3fa0ec-e25b-4fae-a423-67d59b126062

## References
Fix for issue #438 

1. Tried creating a project with a valid name → Works as expected.
2. Tried creating a project with an empty name → Error is shown.

Yes, minor update required to specify project name validation in the documentation.

## Checklist:

### DIFF
--- a/datashuttle/tui/configs.py
+++ b/datashuttle/tui/configs.py
@@ -424,10 +424,15 @@ class ConfigsContent(Container):
         project_name = self.interface.project.cfg.project_name
         # Add validation before proceeding
         if not project_name.strip():
-            self.notify("Project name cannot be empty. Please enter a valid name.", severity="error")
-            print("Project creation blocked due to empty name!")  # Debugging print
+            self.notify(
+                "Project name cannot be empty. Please enter a valid name.",
+                severity="error",
+            )
+            print(
+                "Project creation blocked due to empty name!"
+            )  # Debugging print
             return
-        
+
         for key, value in cfg_kwargs.items():
             saved_val = self.interface.get_configs()[key]
             if key in ["central_path", "local_path"]:
@@ -453,11 +458,15 @@ class ConfigsContent(Container):
         project_name = self.query_one("#configs_name_input").value
         # Add validation before proceeding
         if not project_name.strip():
-            self.notify("Project name cannot be empty. Please enter a valid name.", severity="error")
-            print("Project creation blocked due to empty name!")  # Debugging print
+            self.notify(
+                "Project name cannot be empty. Please enter a valid name.",
+                severity="error",
+            )
+            print(
+                "Project creation blocked due to empty name!"
+            )  # Debugging print
             return
-        
-        
+
         cfg_kwargs = self.get_datashuttle_inputs_from_widgets()
 
         interface = Interface()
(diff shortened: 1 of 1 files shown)


datashuttle/tui/configs.py
```diff
@@ -446,6 +456,17 @@ def setup_configs_for_a_new_project(self) -> None:
         with the new project.
         """
         project_name = self.query_one("#configs_name_input").value
+        # Add validation before proceeding
+        if not project_name.strip():
+            self.notify(
+                "Project name cannot be empty. Please enter a valid name.",
+                severity="error",
+            )
+            print(
```
parharti, line 465: I think instead of using a print statement, it would be more helpful to display this message in a dialog box. That way, users are immediately informed about the issue without needing to check logs.

chandrikarj: Hi @parharti, well your suggestion makes sense, and, the users don't need to check logs and can already see the statement in a dialog box. I had added the print statement to debug the same. You can see it in the video linked under demonstration of the changes. Thanks

datashuttle/tui/configs.py
```diff
@@ -446,6 +456,17 @@ def setup_configs_for_a_new_project(self) -> None:
         with the new project.
         """
         project_name = self.query_one("#configs_name_input").value
+        # Add validation before proceeding
+        if not project_name.strip():
```
JoeZiminski, line 460: Just as a general tip, if you have a piece of code you are using in one place, you can create a function and place the code there, then call it from multiple places. The benefit of this is that if you need to change the code in future, you only have to do it in one place.

JoeZiminski (changes requested):
Hey @chandrikarj thanks a lot for this! This implementation is working very nicely. I think however we can make this change further up the datashuttle call chain, and it will show the error both in the TUI and in the python API.

Under the hood, when the TUI is instantiated the python API is used. In code this would look like:

```
from datashuttle import DataShuttle

project = DataShuttle(")
```

This instantiates this class and sets the project name [here](https://github.com/neuroinformatics-unit/datashuttle/blob/f12971366f917d5fa634f0909c9fa681722bac09/datashuttle/datashuttle_class.py#L108).

If we place a check here, it will raise and error both in the python API and in the TUI for free. I just realized there is already a function here called `_error_on_base_project_name` to check the name does not contain special characters, but it misses the empty case. You could try adding the case here, and I think it will be caught in the python API and TUI. 

Let me know if you have any questions!

chandrikarj:
Thanks for your comments, working on this!

datashuttle/tui/configs.py
parharti: got your point!!

JoeZiminski:
Hey @chandrikarj hope everything is going well, further to our chat would you be be happy for me to to take over this PR (I just checked and I am able to push to the branch). Cheers!

Issue #330, opened 2024-03-15
Disable the 'save' button on configs when widget input is changed.


JoeZiminski:
Current it is a little confusing whether to re-save after SSH connection has been setup in the TUI configs tab. 

It would be cool to have 'Save' disabled until a widget has something enterd in, and then become active.  Would need to check what happens on the project manage tab when a config is changed and the user switches tab, but I think the state is maintained so should be okay.

TheAnkushAgarwal:
Hi @JoeZiminski , I have solved this issue and raise a PR. Please do review it and suggest me if any changes are required.

JoeZiminski:
This issue was developed in #476 and @TheAnkushAgarwal made a nice implementation. In practice, there are quite a lot of edge cases and to cover all fully would require a lot of code changes, as described [here](https://github.com/neuroinformatics-unit/datashuttle/pull/476#pullrequestreview-2669459832). On balance, the benefit of fixing this issue (making it slightly clearer when the configs are filled in) would not be worth the code maintenance and might be confusing in their own right (for example the 'save' button will be enabled / disabled based on filling in various combinations of settings, and users might not be able to intuitively link the status of the save button to the actions they are taking.

Thanks @TheAnkushAgarwal for the PR, this has been very useful!

PR #476, opened 2025-03-07
Enable the 'save' button on configs when widget input is changed

TheAnkushAgarwal:
Before submitting a pull request (PR), please read the [contributing guide](https://github.com/neuroinformatics-unit/.github/blob/main/CONTRIBUTING.md).

Please fill out as much of this template as you can, but if you have any problems or questions, just leave a comment and we will help out :)

## Description

**What is this PR**
- [ yes] Addition of a new feature
Issue #330 solved. Moreover, this is my first PR in any open source organisation. Please do guide me and correct me if something went wrong from my side. 

**Why is this PR needed?**
This PR is needed because I have tried to solve the issue #330 which is there in your repository

**What does this PR do?**
I have implemented the 'Save'  button as disabled until a widget has something entered in, and then become active in config tab.

## References
#330 

Please explain how any new code has been tested, and how you have ensured that no existing functionality has changed.
- I have run pre-commit run -a to test it. 
- I have also done datashuttle launch and tried to manually test the feature

## Is this a breaking change?

If this PR breaks any existing functionality, please explain how and why.
- No.

If any features have changed, or have been added. Please explain how the
documentation has been updated.
- No, I don't think it requires document updation

## Checklist:

- [yes ] The code has been tested locally
- [no ] Tests have been added to cover all new functionality
- [ no] The documentation has been updated to reflect any changes
- [ yes] The code has been formatted with [pre-commit](https://pre-commit.com/)

### DIFF
--- a/datashuttle/tui/configs.py
+++ b/datashuttle/tui/configs.py
@@ -207,9 +207,6 @@ class ConfigsContent(Container):
             id = "#configs_name_input"
             self.query_one(id).tooltip = get_tooltip(id)
 
-            # Register event handler for project name input
-            self.query_one(id).changed_callback = self.on_input_changed
-
             # Assumes 'local_filesystem' is default if no project set.
             assert (
                 self.query_one("#configs_local_filesystem_radiobutton").value
@@ -233,23 +230,6 @@ class ConfigsContent(Container):
         ]:
             self.query_one(id).tooltip = get_tooltip(id)
 
-        # Register change handlers for input fields
-        for input_id in [
-            "#configs_local_path_input",
-            "#configs_central_path_input",
-            "#configs_central_host_id_input",
-            "#configs_central_host_username_input",
-        ]:
-            self.query_one(input_id).changed_callback = self.on_input_changed
-
-    def on_input_changed(self, widget: ClickableInput) -> None:
-        """
-        Handler for input changes to enable the save button
-        when widgets are modified.
-        """
-        self.has_unsaved_changes = True
-        self.query_one("#configs_save_configs_button").disabled = False
-
     def on_radio_set_changed(self, event: RadioSet.Changed) -> None:
         """
         Update the displayed SSH widgets when the `connection_method`
@@ -261,9 +241,6 @@ class ConfigsContent(Container):
         When mode is `No connection`, the `central_path` is cleared and
         disabled.
         """
-        # Mark changes so save button is enabled
-        self.has_unsaved_changes = True
-        self.query_one("#configs_save_configs_button").disabled = False
 
         label = str(event.pressed.label)
         assert label in [
@@ -371,10 +348,6 @@ class ConfigsContent(Container):
             else:
                 self.setup_configs_for_an_existing_project()
 
-            # After saving, disable the save button again
-            self.has_unsaved_changes = False
-            self.query_one("#configs_save_configs_button").disabled = True
-
         elif event.button.id == "configs_setup_ssh_connection_button":
             self.setup_ssh_connection()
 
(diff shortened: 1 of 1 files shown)


datashuttle/tui/configs.py
```diff
@@ -198,6 +207,9 @@ def on_mount(self) -> None:
             id = "#configs_name_input"
             self.query_one(id).tooltip = get_tooltip(id)
 
+            # Register event handler for project name input
+            self.query_one(id).changed_callback = self.on_input_changed
```
JoeZiminski, line 211: Due to textual magic, `self.on_input_changed` will be automatically be called when an input is changed (no need to set it explicitly)

datashuttle/tui/configs.py
```diff
@@ -221,6 +233,23 @@ def on_mount(self) -> None:
...
             self.query_one(id).tooltip = get_tooltip(id)
 
+        # Register change handlers for input fields
+        for input_id in [
+            "#configs_local_path_input",
+            "#configs_central_path_input",
+            "#configs_central_host_id_input",
+            "#configs_central_host_username_input",
+        ]:
+            self.query_one(input_id).changed_callback = self.on_input_changed
```
JoeZiminski, line 243: This can be removed also

datashuttle/tui/configs.py
```diff
@@ -221,6 +233,23 @@ def on_mount(self) -> None:
...
+
+    def on_input_changed(self, widget: ClickableInput) -> None:
+        """
+        Handler for input changes to enable the save button
+        when widgets are modified.
+        """
+        self.has_unsaved_changes = True
```
JoeZiminski, line 250: This variable can be removed, and the button disabled / enabled directly.

datashuttle/tui/configs.py
```diff
@@ -232,6 +261,10 @@ def on_radio_set_changed(self, event: RadioSet.Changed) -> None:
         When mode is `No connection`, the `central_path` is cleared and
         disabled.
         """
+        # Mark changes so save button is enabled
+        self.has_unsaved_changes = True
+        self.query_one("#configs_save_configs_button").disabled = False
```
JoeZiminski, line 266: Can remove from here, only the inputs changing are relevant to be able to save

datashuttle/tui/configs.py
```diff
@@ -338,6 +371,10 @@ def on_button_pressed(self, event: Button.Pressed) -> None:
             else:
                 self.setup_configs_for_an_existing_project()
 
+            # After saving, disable the save button again
+            self.has_unsaved_changes = False
+            self.query_one("#configs_save_configs_button").disabled = True
```
JoeZiminski, line 376: Can remove these lines also

datashuttle/tui/configs.py
```diff
@@ -392,6 +429,10 @@ def handle_input_fill_from_select_directory(
                 path_.as_posix()
             )
 
+        # Enable the save button since we've changed a path
```
JoeZiminski, line 432: Can remove here also as these are automatically triggered by `self.on_input_changed`

JoeZiminski (commented):
Thanks a lot for this @TheAnkushAgarwal this was fun to play around with, I have left some comments above.

Having played around with this, I think the idea of enabling / disabling the save button might not be a good one after all (this is not related to the implementation here, but the idea expressed in the original issue). The problem is is that there are specific combinations of widgets filled it that means saving should be allowed. For example in local-only mode, only the project name and local path must be filled in. However, in ssh mode the all the ssh inputs must be filled in.  Codifying this would take up a lot of effort and maintenance, which I don't think is worth it for the feature. Otherwise, setting save enabled when any input is filled in might be confusing, as the user might think it's possible to save for example after filling in 'project name' but they do also need to fill in 'local path'. As such it is probably easiest to keep the behavior as-is in main, even though it is a little sub-par.

Let me know what you think about this, in general do not worry if this PR is not merged. It is just as useful to explore the solution for an issue in a PR and re-assess it. We can link this PR to the issue and close it, and it will be useful information for future developers.

JoeZiminski:
On balance, the benefits of the functionality added were outweighed by costs in code implementation, due to edge cases that the original issue #330 did not consider. See #330 for more detail, this PR can act as a reference for a nice implementation of this in case the idea is ever revisited.

Issue #450, opened 2024-12-18
Support SSH to windows machine


JoeZiminski:
Currently the ssh key setup assumes the target machine is bash [here](https://github.com/neuroinformatics-unit/datashuttle/blob/1be876e97de531f3c3bd6b616a2e415a1d8481ca/datashuttle/utils/ssh.py#L48)

and fails on windows. The SSH setup will silently fail, and on transfer you will get:

```
2024/12/18 15:29:10 DEBUG : pacer: low level retry 5/10 (error couldn't connect SSH: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain)
```

Background: when setting up passwordles ssh connection, you will get a public and private key. The public key needs to be added to `authorized_keys` in `~/.ssh`. This step is failing on windows. When rclone connects to a server, `sshd` (whihc manages ssh connection) will check the authorized_keysfile for allowed public keys. On windows, `sshd_config` found `C:\ProgramData\ssh` here for me contains a lot of configs related to this behaviour.

In datashuttle, the keys are managed by paramiko and stored in a file in the `USER/.datashuttle/<project_name>/<project_name>_ssh_key` file. The public key can be obtained in code with:

```
key = paramiko.RSAKey.from_private_key_file(project.cfg.ssh_key_path.as_posix())
print(f"{key.get_name()} {key.get_base64()}")
```

As a quick workaround, you can copy these keys to `USER\.ssh\authorized_keys` manually. However, if you are on an admin account, instead `sshd` it will (annoy and confusingly) check `C:\ProgramData\ssh\administrators_authorized_keys` instead.

Some other things to note when setting up the `authorized_key` file:

You can set permissions on the `authorized_key` file (important):
```
# Ensure the .ssh directory is secured
icacls "C:\Users\jzimi\.ssh" /inheritance:r /grant "jzimi:(OI)(CI)F"

# Ensure the authorized_keys file is secured
icacls "C:\Users\jzimi\.ssh\authorized_keys" /inheritance:r /grant "jzimi:(F)"
```

and you may need to set `PubkeyAuthentication yes` in `sshd_config`. 

You can troubleshoot sshing with `ssh -v -i C:\Users\jzimi\.ssh\id_ed25519 jzimi@localhost`

[...]

PR #477, opened 2025-03-07
ADD: Support SSH to Windows Machine

cs7-shrey:
## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**
Fixes Issue #450 : Support SSH to Windows Machine

**What does this PR do?**
Allows storing public key on the server during a SSH to a Windows Machine.
## References
Issue #450 

Minor testing with a windows machine.

**Please explain how any new code has been tested, and how you have ensured that no existing functionality has changed**
Connection and data transfer from a windows machine was tested.

## Testing Approach
- Create / update the SSH config for the windows machine
- Click "Setup SSH Connection" and complete the setup for SSH keys
- Now, to test if the SSH setup was successful, we can test in two ways:
- - a. Try SSHing into the windows machine with the private key created by datashuttle 
(use `ssh -i <path_to_private_key>`) (you may need to remove your own public key from the authorized keys file on windows since SSH will try to use all keys availabe, or you can try SSH with `ssh -v` to inspect the logs and see if your key was accepted.
- - b. Have a project on the windows machine and try transferring data to and from the windows machine 
- If the above two methods work, then it can be assumed that SSH setup was successful

## Is this a breaking change?
No

No

## Checklist:

### DIFF
--- a/datashuttle/utils/ssh.py
+++ b/datashuttle/utils/ssh.py
@@ -48,12 +48,18 @@ def connect_client_core(
 
 
 def is_windows(client: paramiko.SSHClient) -> bool:
+    """
+    Checks the target machine for a Windows OS using two approaches
+        1. execute "ver" : outputs "Microsoft Windows <specific_version>"
+        2. check for the presence of powershell
+    """
     try:
         stdin, stdout, stderr = client.exec_command("ver")
         output = stdout.read().decode().strip().lower()
         if "windows" in output:
             return True
 
+        # rechecking in case the previous check fails for some reason
         stdin, stdout, stderr = client.exec_command(
             "powershell -Command echo 'powershell'"
         )
@@ -105,11 +111,13 @@ def add_public_key_to_central_authorized_keys(
                     f"echo {key.get_name()} {key.get_base64()}"
                     f">> %USERPROFILE%\\.ssh\\authorized_keys"
                 )
+                # setup permissions for .ssh directory
                 client.exec_command(
                     "icacls %USERPROFILE%\\.ssh /inheritance:r /grant %USERNAME%:(OI)(CI)F"
                 )
+                # setup permissions for authorized_keys file
                 client.exec_command(
-                    "icacls %USERPROFILE%\.ssh\\authorized_keys /inheritance:r /grant %USERNAME%:(F)"
+                    "icacls %USERPROFILE%\\.ssh\\authorized_keys /inheritance:r /grant %USERNAME%:(F)"
                 )
 
         else:
--- a/datashuttle/utils/ssh.py
+++ b/datashuttle/utils/ssh.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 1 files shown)


datashuttle/utils/ssh.py
```diff
@@ -61,15 +90,37 @@ def add_public_key_to_central_authorized_keys(
...
+                )
+            else:
+                client.exec_command("mkdir %USERPROFILE%\\.ssh")
+                client.exec_command(
+                    # double >> for concatenate
+                    f"echo {key.get_name()} {key.get_base64()}"
+                    f">> %USERPROFILE%\\.ssh\\authorized_keys"
+                )
+                client.exec_command(
+                    "icacls %USERPROFILE%\\.ssh /inheritance:r /grant %USERNAME%:(OI)(CI)F"
```
JoeZiminski, line 109: Could you add a couple of comments here explaining what these steps do?

JoeZiminski (commented):
Hi @cs7-shrey thanks a lot for this, it's a very clean implementation. I don't have many suggestions, I think this is probably good to go in its current form. 

The main issue here will be testing, to test over a few different machines / windows versions and check everything is working okay, and then automate these tests. This is related to #208 and #100, which cover extending our test cases (some public, but some internal for our institute).

I'll look into this further and get back to you ASAP! We will definitely move forward with this PR, I'm just not certain on the timeline as mentioned on the zulip chat.  But I will try and test it over the next couple of weeks and get back to you here! 

Thanks again! this is a very nice addition.

cs7-shrey:
@JoeZiminski I have made changes as you suggested. I wanted to ask one more thing. The current approach does not change file/directory permissions if the windows user is an admin as I assumed that admin accounts have the already required permissions and restrictions. Do you suggest explicitly setting up permissions for admin users too?

datashuttle/utils/ssh.py
```diff
@@ -61,15 +96,39 @@ def add_public_key_to_central_authorized_keys(
...
         else:
             connect_client_core(client, cfg, password=password)
+        if is_windows(client):
+            if is_windows_user_admin(client):
+                client.exec_command("mkdir C:\\ProgramData\\ssh")
+                client.exec_command(
+                    # double >> for concatenate
+                    f"echo {key.get_name()} {key.get_base64()}"
+                    f">> C:\\ProgramData\\ssh\\administrators_authorized_keys"
+                )
```
cs7-shrey, line 106: is it suggested to add commands to change file permissions for admin users?

JoeZiminski: Good point @cs7-shrey, I think it makes sense to assume file permissions are already granted, as you say. If it turns out that its causing problems in testing, we can add this later if required.

JoeZiminski:
The original case for this turned out to be more prototyping than a real-world case, I think this is unlikely to come up often. Also, in such a situation (unless many people start asking for it) using Windows Subsystem for Linux would probably be the answer. Nonetheless thanks @cs7-shrey this was very useful to see coded up!

Issue #440, opened 2024-10-11
Ensure new version is only build on the correct tag


JoeZiminski:
It is quite a common mistake to forget the leading `v` when creating a new version tag. This is solved with this approach [here](https://github.com/brainglobe/brainglobe-stitch/pull/14#discussion_r1777444270). Implement it for datashuttle.

sumana-2705:
Hello @JoeZiminski I want to work on this issue, can you please assign it to me?

adamltyson:
Hi @sumana-2705 you don't need to be assigned to work on an issue. Just raise a draft PR early and link this issue so it's clear that you're working on it,.

sumana-2705:
Thanks for your reply @adamltyson Sure I will open the draft soon.

sumana-2705:
Hello @TheAnkushAgarwal. Congratulations for your PR 🎉, but please conform once before opening a PR that anyone else are working on the issue for further issues.

adamltyson:
Hi @sumana-2705, I understand the frustration that someone else may be working on the same issue as you. However, as outlined above, the way you can ensure nobody else does this is by raising a draft PR yourself once you have started working on the issue. Unfortunately we can't "reserve" issues based on a GitHub comment, we would need to check in with lots of people before working on issues ourselves.

sumana-2705:
Hello @adamltyson . Thank you for the clarification. I understand the process now and will make sure to open a draft PR as soon as I start working on an issue in the future.

PR #481, opened 2025-03-11
Add v* in workflow tags to solve issue 440

TheAnkushAgarwal:
Before submitting a pull request (PR), please read the [contributing guide](https://github.com/neuroinformatics-unit/.github/blob/main/CONTRIBUTING.md).

Please fill out as much of this template as you can, but if you have any problems or questions, just leave a comment and we will help out :)

## Description

**What is this PR**

- [ yes] Bug fix
- [ no] Addition of a new feature
- [ no] Other

**Why is this PR needed?**
Closes: #440 
It is quite a common mistake to forget the leading v when creating a new version tag. So , I have solved this issue (#440) by adding 'v**' in tags of workflow. 

**What does this PR do?**
If the "v" is omitted, the system might not recognize the version correctly, potentially causing issues during the build or deployment process. Some CI/CD pipelines or build scripts are set to enforce this naming convention, meaning if the version tag doesn't start with a "v", the build might fail. However, some systems might still allow the build to proceed without it, leading to inconsistencies.

## References
Issue #440 

Please reference any existing issues/PRs that relate to this PR.

- by running pre-commit run -a
- by manually testing the workflow.

Please explain how any new code has been tested, and how you have ensured that no existing functionality has changed.

## Is this a breaking change?
- no

If this PR breaks any existing functionality, please explain how and why.

-no

If any features have changed, or have been added. Please explain how the
documentation has been updated.

## Checklist:

- [ yes] The code has been tested locally
- [yes ] Tests have been added to cover all new functionality
- [ yes] The documentation has been updated to reflect any changes
- [ yes] The code has been formatted with [pre-commit](https://pre-commit.com/)

### DIFF
--- a/.github/workflows/code_test_and_deploy.yml
+++ b/.github/workflows/code_test_and_deploy.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - main
     tags:
-      - 'v**'
+      - 'v*'
   pull_request:
   schedule:
     # Weekly cron job at 12:00 AM UTC on Mondays.
--- a/.github/workflows/code_test_and_deploy.yml
+++ b/.github/workflows/code_test_and_deploy.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - main
     tags:
-      - '*'
+      - 'v**'
   pull_request:
   schedule:
     # Weekly cron job at 12:00 AM UTC on Mondays.
--- a/.github/workflows/docs_build_and_deploy.yml
+++ b/.github/workflows/docs_build_and_deploy.yml
@@ -10,7 +10,7 @@ on:
     branches:
       - main
     tags:
-      - '*'
+      - 'v**'
   pull_request:
   workflow_dispatch:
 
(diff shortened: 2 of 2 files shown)


.github/workflows/code_test_and_deploy.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches:
       - main
     tags:
-      - '*'
+      - 'v**'
```
JoeZiminski, line 8: In this case, we will not have multi-segment tags (we use semantic versioning only), as such 'v*' will suffice here (and the other change).

JoeZiminski (changes requested):
Thanks @TheAnkushAgarwal! This will come in very useful for ensuring that incorrect-formatted tags are not accidentally pushed. I made one small comment, then this is good to go!

.github/workflows/docs_build_and_deploy.yml
```diff
@@ -10,7 +10,7 @@ on:
     branches:
       - main
     tags:
-      - '*'
+      - 'v**'
```
JoeZiminski, line 13: thanks @TheAnkushAgarwal, and here too!

JoeZiminski (changes requested):

JoeZiminski (approved):
Thanks @TheAnkushAgarwal LGTM

Issue #409, opened 2024-07-02
Add suggest next sub / ses including remote project to TUI


JoeZiminski:
Currently for speed reasons, through the TUI  suggesting the next subject or session is done with reference to the local project only (not remote). It would be nice to expose this (it is just an argument on `get_next_sub` / `get_next_ses`). However, it should be option as when using SSH (and, in some cases mounted drive) it will be slow

JoeZiminski:
A nice suggestion related to this is to make the example placeholders in the sub / ses input boxes (e.g. `sub-001`) actually be the suggested next sub / session. When we can be sure that it is indeed the correct next sub / ses based on the entire project (also see #373).

cs7-shrey:
Hi @JoeZiminski I want to work on this issue. I noticed that searching both central and local repositories takes a bit longer.  
Do you suggest adding a checkbox? So that users can enable searching on both central and local repos. 
And a loading indicator to indicate "processing" meanwhile the TUI generates suggestion for next sub / ses 
[again, the workers issue]

How do you suggest the UI be?

JoeZiminski:
Hey @cs7-shrey thanks for looking into this, indeed searching the remote takes quite a long time (especially when connected via SSH) and so it makes sense to turn this on with a checkbox as you suggest. I think a checkout above 'Bypass Validation' in the 'Create Folders',  'Settings' window. It could be called something like 'Search central for next sub / ses" and have a tooltip that explains that this may lead to reduced performance.

It would be cool to also do this in a worker! I'm not sure if its best to only use a worker when searching remotely, or if it will cause some code duplication. I think though that when searching locally, the 'processing' screen would flash up momentarrily and look a bit strange. So I think it makes sense to start implementing `local_only=False` condition in a worker. Then we can look to reduce the code duplication as much as possible. 

Thanks for looking into this, please let me know if you have any questions!

PR #484, opened 2025-03-13
Suggest next sub ses remote

cs7-shrey:
## Description

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**
Solves issue #409 

**What does this PR do?**
Allows users to search both local and remote repositories for next sub / ses suggestion
## References
Issues #409 

Manually tested on local and remote projects for next sub / ses suggestions. Testing still in progress.

## Is this a breaking change?
No

No

## Checklist:


cs7-shrey:
@JoeZiminski please review the current approach for this feature. 

To test this locally, please consider creating a new project since this PR adds a new key to the persistent settings.

or you can change the persistent_settings.yaml file, you'll need to add `suggest_next_sub_ses_local_only: false` above `name_templates: `

JoeZiminski:
Hey @cs7-shrey thanks for this, its looking great! I just had a quick look to play around with it on a new project and it looks like it's working perfectly. I will try and review ASAP.

The only thing things to add will be a test to ensure it's working going forward, and documentation. You can build the documentation locally with [these instructions](https://datashuttle.neuroinformatics.dev/pages/community/contributing.html#contributing-documentation). Feel free to have a look through and let me know if you see anywhere it would make sense to add a note on this button, I think it will be useful to highlight it in the docs for users. 

For tests, the approach would be similar to [this test](test_get_next_sub_and_ses_no_template) but instead of creating folders locally, it would create folder centrally and then check these are detected (similar to [here](https://github.com/neuroinformatics-unit/datashuttle/blob/16ac2ef559edcaccb0f5cd07e88cb05e7a2e49a0/tests/tests_integration/test_create_folders.py#L321)). These tests are more extensive, in this case I think it would make sense to do something like:

1) make a sub-001 locally and a sub-001 and sub-002 centrally.
2) check with the checkbox off sub-002 is suggested
3) turn the checkbox on and check sub-003 is selected.

This will be sufficient to check that the `local_only` flag is passed.

Another approach is to monkeypatch the `datashuttle.get_next_sub()` and `get_next_ses()` function similar to [here](https://github.com/neuroinformatics-unit/datashuttle/blob/16ac2ef559edcaccb0f5cd07e88cb05e7a2e49a0/tests/tests_tui/test_tui_directorytree.py#L155) and just check the underlying functions are called with `local_only=True`. On balance, as we will have to set up a lot of the TUI just to trigger this monkeypatched function, we might as well just check the output on the TUI side also.

Thanks again for this nice addition and please let me know if you have any questions!

JoeZiminski:
BTW I forgot to say, there is this function [here](https://github.com/neuroinformatics-unit/datashuttle/blob/8b9c118cfc5d02263ab6bdb7c9cbfc35f8eb149c/datashuttle/datashuttle_class.py#L1523) which ensures persistent settings are backwards compatible. The new key can be added to it, the idea is to insert new keys that are not found in the dictionary (i.e. if they are from old versions).

datashuttle/tui/tabs/create_folders.py
```diff
@@ -153,10 +157,26 @@ def on_clickable_input_clicked(
 
         prefix: Prefix = "sub" if "subject" in input_id else "ses"
 
-        if event.ctrl:
-            self.fill_input_with_template(prefix, input_id)
-        else:
-            self.fill_input_with_next_sub_or_ses_template(prefix, input_id)
+        async def _on_clickable_input_clicked():
```
JoeZiminski, line 160: This is nice, it works really nicely, as a slgihtly refactoring, I think 

```
            if event.ctrl:
                self.fill_input_with_template(prefix, input_id)
```
can be moved outside of the async function e.g.
```
    if event.ctrl:
        self.fill_input_with_template(prefix, input_id)
    else:
        async def ....
```

datashuttle/tui/tabs/create_folders.py
```diff
@@ -153,10 +157,26 @@ def on_clickable_input_clicked(
...
+            if event.ctrl:
+                self.fill_input_with_template(prefix, input_id)
+            else:
+                input_box = self.query_one(f"#{input_id}")
```
JoeZiminski, line 164: I think this is a cool approach, I like the spinner widget I did not know about that! In this case I have three thoughts:

- I'm not sure if you get this, but for me double-clicking on the input makes the directory tree flash, I'm not 100% sure why this is happening 🤔 
- The spinner display is quite small inside the input, I didn't actually notice it for the first few times clicking
- It is very nice option that the GUI remains usable while the connection is happening. However, in this case I think it makes sense to tightly control what the user can do during this period, at least for now. The reason is if the connection takes a long time, the user might do some unexpected things. For example, they might close the project, open another project, and so something else. All the while the other task would be running in the background. This might lead to some unexpected behavior and could feasibly crash the software.

Considering the above points, I wonder if having a dialog pop-up similar to the transfer tab. It could say something like 'Searching central for subjects and sessions" with a progress bar. It would make sense down the line to merge this with transfer, but because the transfer is quite complex, for now these could be two different widgets, We could create an issue to merge these later. Let me know what you think of his idea.

datashuttle/tui/tabs/create_folders.py
```diff
@@ -296,10 +316,17 @@ def fill_input_with_next_sub_or_ses_template(
             "top_level_folder_select"
         ]["create_tab"]
 
+        def show_error_dialog(output: str) -> None:
```
JoeZiminski, line 319: I like the idea of encapsulating this behavior in a  function, both to avoid duplication and keep the code clean. In general I try to avoid inner-functions, because they inherit the scope of the outer function which can lead to some confusing behavior. This is a personal thing but in general it is nice to rely on the notion that only inputs to a function are available in it's scope, otherwise it could have unintended side-effects if variables from the outer-scope are modified. They are useful if you explicitly want to encapsulate the scope of the outer-function, otherwise it can be worth thinking about alternatives. 

In this case, I wonder if a mainwindow function `show_modal_error_dialog_from_thread()` can be defined, that has the body of `show_error_dialog`. I think this could also be used in your other PR (we can combine these when both are merged) and in general across the codebase. Let me know what you think!

JoeZiminski (changes requested):
Hey @cs7-shrey thanks a lot for this! Looking good, please see some points of discussion in the review. Let me know if you have any questions about anything.

JoeZiminski:
Hey @cs7-shrey sorry for all the messages, just to let you know in #491 the argument "local_only" was changed to "include_central" and is now by default `False` (before it was default `True` but always set to `False` in the TUI.) 

I will merge '491 shortly, I think for this PR it will just entail renaming the arguments and switching the bool value (as it is now reversed). The easiest way to handle this is probably to 'rebase' this PR onto the new version of main. Once #491 is merged into main (I'll send you a ping) you can do:

1) Go to [your fork main branch](https://github.com/cs7-shrey/datashuttle/tree/main) and click the 'sync' button so that your fork main is up to date with `datashutle` main.

```
git checkout main
git pull
git checkout suggest_next_sub_ses_remote
git rebase -i main
```
This will start an interactive rebase (see [this guide](https://hackernoon.com/beginners-guide-to-interactive-rebasing-346a3f9c3a6d) for detailed information) on running and handling conflicts. Let me know if you have any questions!

cs7-shrey:
Cool, no problem! It will just require some minor changes.

JoeZiminski:
Just FYI #491  is now merged!

datashuttle/tui/tabs/create_folders.py
cs7-shrey: I think I know why the directory tree flashes. On double clicking, when a click is registered, the input box comes in focus. For suggesting the next sub/ses, the input box would be disabled (and goes out of focus) so the directory tree / the main screen gets back the focus due to which it flashes. 

About the dialog popup, I think it would be the best way to restrict user activity during searching central for sub/ses. About the spinner, I thought from a UX perspective that it would be good to keep the animations to the input box instead of a full fledged popup (as seen in many websites) for something that suggests the folder names in an input box. But for larger projects, this processing would take longer and longer time and then a dialog popup would make sense. 

For conclusion, as you have used datashuttle a lot more than I have and interacted with users using datashuttle, you would be in a better place to decide how the final UI should look like. Please let me know what you think would be the best.

cs7-shrey: > I think I know why the directory tree flashes. On double clicking, when a click is registered, the input box comes in focus. For suggesting the next sub/ses, the input box would be disabled (and goes out of focus) so the directory tree / the main screen gets back the focus due to which it flashes.
> 
> About the dialog popup, I think it would be the best way to restrict user activity during searching central for sub/ses. About the spinner, I thought from a UX perspective that it would be good to keep the animations to the input box instead of a full fledged popup (as seen in many websites) for something that suggests the folder names in an input box. But for larger projects, this processing would take longer and longer time and then a dialog popup would make sense.
> 
> For conclusion, as you have used datashuttle a lot more than I have and interacted with users using datashuttle, you would be in a better place to decide how the final UI should look like. Please let me know what you think would be the best.

Hey @JoeZiminski I added a comment please have a look. I think github doesn't notify about this.

cs7-shrey:
Hey @JoeZiminski, I added a [comment](https://github.com/neuroinformatics-unit/datashuttle/pull/484#discussion_r2001340171), please have a look! I thing github doesn't notify about that

cs7-shrey:
Hey @JoeZiminski I think something caused the tests to run I'm not sure why. Please disable them for now as it is still in development. Also, I had to force-push some changes as the remote branch wasn't accepting changes from local branch and a `git pull` to merge remote to local started showing existing changes (done on some previous PRs that were merged) as new changes in my branch in the pull request.

sumana-2705:
Hello @cs7-shrey,  

This problem is very common and can be quite irritating, I face it sometimes too. The issue is likely due to a divergence between the local and remote branches, possibly caused by a force-push in the repository history. When it occurs, I usually sync my forked repo (remote) and then run the following commands in local branch:  

```bash
git fetch origin
git status
```
I guess this should solve the issue. Additionally, you shouldn't need to force-push the commits after this. Let me know if you need any further help! 😊

cs7-shrey:
Hey @JoeZiminski, I've modified the code to show a popup on searching remote for sub/ses. Let me know what you think of the current approach.

datashuttle/tui/tooltips.py
```diff
@@ -121,6 +121,11 @@ def get_tooltip(id: str) -> str:
     elif id == "#create_folders_settings_toplevel_select":
         tooltip = "The top-level-folder to create folders in."
 
+    elif id == "#suggest_next_sub_ses_remote":
+        tooltip = (
+            "Search the remote project folder for folder suggestions. "
```
JoeZiminski, line 126: Just to add a little more detail, we could say `Search the central project folder when suggesting the next subject or session. May be slower compared than searching local only.`

JoeZiminski (changes requested):
Hey @cs7-shrey this looks great! Works really well, I made some small suggestions on things like naming / some minor refactoring's but nothing substantial. 

We'll have to add some tests for this, I think the two things to test are:

1) That the status of the button is remembered across sessions
2) that the `include_central` tag is propagated down from the tui to the underlying datashuttle functions (through interface)

In contrast to my previous message, I think it would actually be easier to just monkeypatch `get_next_sub` and `get_next_ses` in the tests. [here](https://github.com/neuroinformatics-unit/datashuttle/blob/16ac2ef559edcaccb0f5cd07e88cb05e7a2e49a0/tests/tests_tui/test_tui_create_folders.py#L474) is an example of the set up for testing connect text ses / sub in the TUI. We can monkeypatch the datashuttle `get_next_sub` or `get_next_ses` as is done [here](https://github.com/neuroinformatics-unit/datashuttle/blob/16ac2ef559edcaccb0f5cd07e88cb05e7a2e49a0/tests/tests_tui/test_tui_datatypes.py#L131) with the `mocker` package. In the tests, we can click the input boxes, and check the underlying function is called correctly (with `include_central=False`). Then we can turn on the `Suggest next...` checkbox e.g. [here](https://github.com/neuroinformatics-unit/datashuttle/blob/16ac2ef559edcaccb0f5cd07e88cb05e7a2e49a0/tests/tests_tui/test_tui_create_folders.py#L262) for an example of switching the other checkboxes).

As for testing the setting is maintained across sessions, there is currently a full test of bypass validation [here](https://github.com/neuroinformatics-unit/datashuttle/blob/b82cf9ee62f0a69b137b2954f31245789050a45c/tests/tests_tui/test_tui_widgets_and_defaults.py#L652). I think we can repeat this for the new checkbox and settings.  When writing this test, it might be that it is not worth the duplication and there may be an opportunity to centralise this, or maybe the tests are overkill. Feel free to say as such, as they are quite extensive.

As for the documentation, we can add a small section explaining this in the `Create Folders` page. However, the documentation are undergoing a big refactor so we can wait until #499 is merged before proceeding.

JoeZiminski:
Hey @cs7-shrey sorry I forgot one thing, would it be a pain to make a test case that triggers `dismiss_popup_and_show_modal_error_dialog_from_thread`? It is unlikely to cause any problems but it would be good to have the automated testing check it at least once (just incase something e.g. a textual update breaks it in future)

cs7-shrey:
Hey @JoeZiminski, thanks a lot for the textual update! I'm no longer having those unexpected errors. There was one particular error that was coming up in testing the error popup. After clearing the subject input box, a double click on session input box was throwing an error

It was because of repeated clicks being done in a short period of time due to which a click on subject box and the 1st click on sesion box were together being assumed as a double click and therefore the double click on session box couldn't be completed. 

So, I added a new decorator for that purpose. Please have a look when you find time.

JoeZiminski:
Hey @cs7-shrey this is a very elegant solution, I could replicate this bug manually, great job on the tests to find this subtle bug.

I would suggest one change so that the two existing functions can be combined. We can introduce a class that handles all kinds of click-related infomration, then if we want to extend it in future we can avoid adding more attributes to the relevant classes. It is very useful to ensure the same widget is clicked twice generally, so this is acheived here by extending your solution to the only other use case in datashuttle (directory tree). If we want to support more widgets this will require extendign this function which isn't ideal, but I think the other benefits outway the downside. What do you think?


It requires replacing all `prev_click_time ` or `prev_click_input_id` with
`self.click_info = ClickInfo()`

```
from __future__ import annotations

from functools import wraps
from time import monotonic

# -----------------------------------------------------------------------------
# Double-click decorator
# -----------------------------------------------------------------------------



from time import monotonic
from functools import wraps


class ClickInfo:
    """
    A class to hold click-info to checking
    double clicks are within the time threshold
    and match the widget id.
    """
    def __init__(self):

        self.prev_click_time = 0.0
        self.prev_click_widget_id = ""


def require_double_click(func):
    """
    A decorator that calls the decorated function
    on a double click, otherwise will not do anything.

    Requires the first argument (`self` on the class) to
    have the attribute `click_info`). Any class holding a widget
    that supports double-clicking must have the attribute
    self.click_info = ClickInfo()

    The first (non-self) argument depends on the decorated function,
    which is usually widget-specific. Unfortunately, these must be
    supported on a case-by-case bases and extended when required.
    """

    @wraps(func)
    def wrapper(*args, **kwargs):
        parent_class = args[0]

        assert hasattr(parent_class, "click_info"), (
            "Decorator must be used on class method where the class as "
            "the attribute `self.click_info = ClickInfo()`."
        )

        click_time = monotonic()
        event = args[1]

        if hasattr(event, "input"):
            id = event.input.id
        elif hasattr(event, "node"):
            id = event.node.tree.id
        else:
            raise RuntimeError("The message type for the widget you are trying to"
                               "register clicks on is not supported. Add it to the decorator.")
        if (
            click_time - parent_class.click_info.prev_click_time < 0.5
            and id == parent_class.click_info.prev_click_widget_id
        ):
            parent_class.click_info.prev_click_time = click_time
            parent_class.click_info.prev_click_widget_id = id
            return func(*args, **kwargs)

        parent_class.click_info.prev_click_time = click_time
        parent_class.click_info.prev_click_widget_id = id

    return wrapper

```

JoeZiminski (changes requested):
Hey @cs7-shrey this looks great, just the one suggestion in the other comment and I think we're good to go

cs7-shrey:
Hey @JoeZiminski I'm glad this work was useful. For the node and event confusion, I just ran the code in a debugger (to know the type of the argument), I found it to be `DirectoryTree.DirectorySelected` and `DirectoryTree.FileSelected` (for `on_directory_tree_directory_selected` and `on_directory_tree_directory_selected` respectively)

I also found [this](https://github.com/Textualize/textual/blob/51e00c774230fbeeea1c8c39842fd59b6d29a5f0/examples/code_browser.py#L57) example in the textual source code.

cs7-shrey:
Hehe 😄
Np

PR #495, opened 2025-03-20
[ENH] Show Pop-up Warning for Existing Folder Creation in TUI

sumana-2705:
## Description

Currently, when users attempt to create folders with names that already exist, the TUI silently passes the instruction without informing them. This PR introduces a pop-up notification that alerts users when they try to create a folder that already exists. This ensures better user feedback and prevents confusion, especially when working with multiple nested folders.

**What is this PR**

- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**
- Ensures users are informed about existing folders.
- Prevents silent failures and enhances usability.

**What does this PR do?**

https://github.com/user-attachments/assets/6cde8e45-2f88-450c-af8e-a434bb4ff333

## References
There is no issue linked to this PR, I added this feature based on my observations.

- Tested by attempting to create folders with names of existing folders.
- Verified that the pop-up appears with the correct folder path.
- Checked that the directory tree remains up to date after the operation.

## Is this a breaking change?

No

It is an enhancement, so not required

## Checklist:

### DIFF
--- a/.github/workflows/code_test_and_deploy.yml
+++ b/.github/workflows/code_test_and_deploy.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - main
     tags:
-      - '*'
+      - 'v*'
   pull_request:
   schedule:
     # Weekly cron job at 12:00 AM UTC on Mondays.
--- a/.github/workflows/docs_build_and_deploy.yml
+++ b/.github/workflows/docs_build_and_deploy.yml
@@ -10,7 +10,7 @@ on:
     branches:
       - main
     tags:
-      - '*'
+      - 'v*'
   pull_request:
   workflow_dispatch:
 
--- a/datashuttle/configs/config_class.py
+++ b/datashuttle/configs/config_class.py
@@ -240,7 +240,9 @@ class Configs(UserDict):
 
         self.logging_path = self.make_and_get_logging_path()
 
-    def make_and_get_logging_path(self) -> Path:
+    def make_and_get_logging_path(
+        self,
+    ) -> Path:
         """
         Build (and create if does not exist) the path where
         logs are stored.
--- a/datashuttle/datashuttle_class.py
+++ b/datashuttle/datashuttle_class.py
@@ -34,6 +34,7 @@ from datashuttle.configs import (
     load_configs,
 )
 from datashuttle.configs.config_class import Configs
+from datashuttle.datashuttle_functions import _format_top_level_folder
 from datashuttle.utils import (
     ds_logger,
     folders,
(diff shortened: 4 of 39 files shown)


sumana-2705:
Hello @JoeZiminski,

I believe this enhancement is necessary to improve user experience. Please review this PR and let me know your feedback. Looking forward to your thoughts!

JoeZiminski:
Hi @sumana-2705 thanks for your initiative in opening this PR, this is a really nice idea with a clean implementation. We thought about adding something like this, but ran into some problems. For example. if you have a `sub-001/ses-001` and then try and make `sub-001/ses-002`, do you get an error? Similarly, if you have `sub-001/ses-001/behav` and then try `sub-001/ses-001/ephys`? It would be possible to add new session or datatype folders in this way. I found this was a problem in my implementation, and dealing with it led to a lot of complexity, even if it is possible. As such we did not proceed with this, even though it is nice from a UI perspective. Let me know if you find this issue also.

sumana-2705:
> if you have a `sub-001/ses-001` and then try and make `sub-001/ses-002`, do you get an error?

Hello @JoeZiminski, this is not giving error with current implementation. The other one is not implemented correctly because I have not considered the third level folders in the tree, I will correct it and updtate the code. Thank you for the suggestion

sumana-2705:
I have corrected the code to check datatype folders too. Here is the behaviour of the TUI after changes:


https://github.com/user-attachments/assets/8d040738-fdba-4bc5-a15d-27629771b541

JoeZiminski:
Thanks @sumana-2705! That is very nice it is flexible and does not raise an error if parent folders already exist. I will have to think about this, because for the UI it makes a lot of sense, but for the python API it might cause problems. In general, functionality is implemented at the python API level (e.g. [here](https://github.com/neuroinformatics-unit/datashuttle/blob/ecbfa2a29710da5227ccb0f809e4b6bf3ef4de9e/datashuttle/utils/folders.py#L31) and then the TUI wraps this functionality. If an error occurs at the python API level, it is propagated to the TUI with the `show_modal_error_dialog` method. So at some point the implemented would be moved to the `folders.py` file linked above. 

Please feel free to play around with this implementation more, I am not 100% sure if will be merged, but it will still be very useful to have this as a reference implementation. The positives of this are great, it will immediately warn someone if they have made a mistake and are trying to create an already-existing folder. The downside is that it may make the python API a little less flexible, and could cause problems in the real world when people are running scripts at they may want to silently skip folders that already exist. Because we are just dealing with folders, there _should_ be an overwrite warning when a user tries to save their raw data.

That been said, a second line of defense would be valuable. It may make sense to have this as default and add an option to turn it off. If you wanted to explore further, implementing this at the python API level would be great and give us an indication if it will cause problems (e..g in tested cases). Thanks a lot for this idea!

sumana-2705:
Hello @JoeZiminski 

Thank you for your valuable feedback. You are absolutely right that implementing this at the API level could lead to errors. The changes I made were focused solely on the TUI without considering the API impact. I truly appreciate your suggestion to provide an option rather than enforcing this behavior, it’s a great idea! However, given the existing layout, I feel that adding another feature, especially an optional one, might clutter the interface. The TUI already displays a lot of information, and incorporating this change may not be the best fit.

That said, working on this PR has been a great learning experience for me, in understanding TUI and file handling. I really appreciate your insights throughout this process. I believe it would be best to close this PR now. I sincerely apologize if I caused any inconvenience with my limited knowledge. Thank you again for your time and guidance!

JoeZiminski:
Hi @sumana-2705, okay great that makes sense, this has been very valuable on our end, to see how this works in practice. If users feed back that this would be a valuable addition, this PR will make a great launch-pad from which to implement the full functionality. Thanks!

Issue #323, opened 2024-03-04
Add tests for renaming folder / file on directorytree


JoeZiminski:
Currently the CTRL+N shortcut is untested.

JoeZiminski:
Also add test for 'overwrite folders' in TUI

JoeZiminski:
Validation should also be added at this level

PR #492, opened 2025-03-17
Add validation to TUI.

JoeZiminski:
## Description

This PR adds validation functionality to the terminal user interface (TUI). In the TUI there are now two new screens:

- Validate Project at Path (wraps `quick_validate_project`) and allows validation of any project at a given filepath.
- 'Validate' tab at the project manager screen, which performs validation of existing projects (wraps `project.validate_project`).

Under the hood there is a central `ValidateContent` container which holds all functionality and is shared between the two screens (with minor changes in display) (similar to configs).

**How to test**

The quick-validate project documentation page has also been updated (see [here](https://datashuttle.neuroinformatics.dev/pages/community/contributing.html#working-with-the-documentation) for building the documentation) and tests are added.

To test, you can run ``datashuttle launch`` in a terminal (see [Choosing a Terminal](https://datashuttle.neuroinformatics.dev/pages/how_tos/choose-a-terminal.html) for the best terminals to run in) in a conda environment that contains datashuttle. You can make a fake [NeuroBlueprint](https://neuroblueprint.neuroinformatics.dev/latest/index.html) project and use datashuttle to validate it through the user interface.

**What is this PR**

- [ ] Bug fix
- [X ] Addition of a new feature
- [ ] Other

**Why is this PR needed?**

Python API validate was not available in the TUI

## How has this PR been tested?

New tests added that check validation results are displayed on the screen and that all settings are faithfully passed down to the underlying functions.

## Is this a breaking change?

No

## Does this PR require an update to the documentation?

Yes, Added. Currently the validation on the project manager window is not fully documented, but this will be added in a follow-up PR as it requires some documentation refactoring.

## Checklist:

- [X ] The code has been tested locally
- [ X] Tests have been added to cover all new functionality
- [X ] The documentation has been updated to reflect any changes
- [X ] The code has been formatted with [pre-commit](https://pre-commit.com/)

PR #496, opened 2025-03-20
Add Tests for Renaming File/Folder on directorytree

cs7-shrey:
## Description

**What is this PR**

- [ ] Bug fix
- [x] Other

**Why is this PR needed?**
Addresses #323 

**What does this PR do?**
Adds tests for "ctrl+n" shortcut on directorytree for renaming files and folders.

## References
Issue #323 

Tests were run locally.

## Is this a breaking change?
No

No

## Checklist:

### DIFF
--- a/tests/tests_tui/test_tui_directorytree.py
+++ b/tests/tests_tui/test_tui_directorytree.py
@@ -245,3 +245,68 @@ class TestTuiCreateDirectoryTree(TuiBase):
             )
 
             await pilot.pause()
+
+    @pytest.mark.asyncio
+    async def test_create_folders_directorytree_rename(
+        self, setup_project_paths
+    ):
+        tmp_config_path, tmp_path, project_name = setup_project_paths.values()
+
+        rawdata_path = tmp_path / "local" / project_name / "rawdata"
+
+        app = TuiApp()
+        async with app.run_test(size=self.tui_size()) as pilot:
+
+            # Set up the 'create tab' with loaded nodes
+            await self.setup_existing_project_create_tab_filled_sub_and_ses(
+                pilot, project_name, create_folders=True
+            )
+
+            await self.reload_tree_nodes(
+                pilot, "#create_folders_directorytree", 4
+            )
+
+            # rename the subject folder
+            await self.hover_and_press_tree(
+                pilot,
+                "#create_folders_directorytree",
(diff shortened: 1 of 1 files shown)


cs7-shrey:
Hey @JoeZiminski, I'd added a test to cover renaming on the directorytree using "ctrl+n" shortcut. Would love to know how we can make this more comprehensive. 

You also mentioned, "Adding test for 'overwrite folders' " and "Adding validation" in #323 
By "adding validation" do you mean adding validation to check for the new name of the sub/ses folder as I can see that renaming folders can easily bypass validation.

And can you please provide a little more context to what "adding test for 'overwrite folders' " means

tests/tests_tui/test_tui_directorytree.py
```diff
@@ -245,3 +245,68 @@ def set_signal_to_path(path_):
...
+    @pytest.mark.asyncio
+    async def test_create_folders_directorytree_rename(
+        self, setup_project_paths
+    ):
+        tmp_config_path, tmp_path, project_name = setup_project_paths.values()
+
+        rawdata_path = tmp_path / "local" / project_name / "rawdata"
+
+        app = TuiApp()
+        async with app.run_test(size=self.tui_size()) as pilot:
```
JoeZiminski, line 258: Nice! This test looks good, one additional suggestion is to check the file is actually changed on disk correctly also. For example, that after renaming there is now no folder called `sub-001`, but there is called `sub-002` for example. Otherwise this is good to go!

JoeZiminski (changes requested):

JoeZiminski:
Hey @cs7-shrey good question, the 'overwrite folder' is old I think and I deleted it, apologies for any confusion. The note on validation was an idea to validate at this stage also, so for example if you tried to rename a folder `sub-001` to `subx-002` it would throw an error. However, on balance I think this would be overkill, and it is nice to allow the user to rename any folder to whatever they like. At this stage they might want to rename non-neuroblueprint folders and it might get confusing. Now the validation is better ingrained into the TUI (#492) any issues should be quickly flagged by them validating. As such I deleted those comments and linked the issue to this PR. Cheers!

JoeZiminski (approved):
This is great, thanks @cs7-shrey!

PR #501, opened 2025-03-24
Added support for custom tags in template input with regex conversion

### LINKED ISSUE #351: Simplify `Name Templates` regexp with custom keywords or a convenience buttons for common use cases
Currently the `Name Templates` feature that allows definition of custom templates to validate against uses straight regexp syntax.

e.g. `sub-\d\d\d_id-.?.?.?.?.?.?_date-.*`

which is obviously pretty confusing.

Two options are 
1) to add in custom tags that are clearer. However it may be hard to tell intended text vs non-intended text (e.g. we use `D` for digit, but what is a `D` is actually in the name).
2) Probably easier is in the TUI to have convenience buttons like 'Any Digit' that can be pressed to automatically fill in the template with `\d`. This makes it easy to make templates in the TUI as well as teach the regexp syntax.

sumana-2705:
## Description

**What is this PR**
- [ ] Bug fix
- [ ] Other

**Why is this PR needed?**
This PR adds support for custom tags in template inputs following the first suggested option in the issue (#351), allowing users to define and use predefined placeholders (`<ANY_DIGIT>`, `<ANY_STRING>` and `ANY_CHAR`) that automatically convert into corresponding regex patterns. This improves flexibility when defining naming conventions.

**What does this PR do?**
- Implements a `convert_custom_tags_to_regex` function to replace custom tags with regex patterns.
- Integrates this function into `set_name_templates` to ensure correct transformation of input values.
- Ensured that users can enter name templates in regex format without encountering any errors.

## References
closes #351

Tested Locally

https://github.com/user-attachments/assets/4d1fa3e9-036c-492e-9c1e-f715cb5f10d6

## Is this a breaking change?
No

Yes, documentation need to be updated when this addition works fine.

## Checklist:

- [ ] The code has been tested locally

### DIFF
--- a/datashuttle/datashuttle_class.py
+++ b/datashuttle/datashuttle_class.py
@@ -1179,6 +1179,21 @@ class DataShuttle:
         settings = self._load_persistent_settings()
         return settings["name_templates"]
 
+    def convert_custom_tags_to_regex(self, template: str) -> str:
+        """
+        Convert custom tags in the template to their corresponding regex patterns.
+        """
+        CUSTOM_TAGS = {
+            "<ANY_DIGIT>": r"\d",  # Matches a single digit (0-9)
+            "<ANY_CHAR>": r".?",  # Matches any single character
+            "<ANY_STRING>": r".*",  # Matches any sequence of characters
+        }
+
+        for tag, regex in CUSTOM_TAGS.items():
+            template = template.replace(tag, regex)
+
+        return template
+
     def set_name_templates(self, new_name_templates: Dict) -> None:
         """
         Update the persistent settings with new name templates.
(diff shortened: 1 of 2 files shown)


sumana-2705:
Hello @JoeZiminski,

I came across an issue while working with the name templates. The [documentation](https://datashuttle.neuroinformatics.dev/pages/how_tos/use-name-templates.html#name-template-placeholders) states that `.?` and `.*` should allow any type of character, but I encountered an error when using them. Based on the observed behavior, I wanted to clarify whether this restriction is intentional (allowing only integers) and the documentation needs an update, or if this is a bug.

If it is indeed a bug, I’d be happy to address it in this PR itself. Looking forward to your guidance on this!

adamltyson:
Hi @sumana-2705, just FYI @JoeZiminski is away, so it may be some time until someone can get back to you.

datashuttle/datashuttle_class.py
```diff
@@ -1179,6 +1179,21 @@ def get_name_templates(self) -> Dict:
         settings = self._load_persistent_settings()
         return settings["name_templates"]
 
+    def convert_custom_tags_to_regex(self, template: str) -> str:
```
JoeZiminski, line 1182: I think this function can be moved for `format.py`

JoeZiminski (changes requested):
(main comments below)

adamltyson:
>  But I think the former would come at the expense of clarity and the latter would clutter the interface too much. @adamltyson what do you think?

I think I agree. We can't just add loads of buttons for every possible option.

sumana-2705:
Thanks for your review @JoeZiminski I will work on it.

JoeZiminski:
Hey @sumana-2705 thanks a lot for this. I think because of the uncertainty on the final implementation of #351, I will link to this PR in the issue and close this PR. The issue will remain open and this will probably be revisited at some point down the line, but is not an immediate priority. Thanks again!

PR #515, opened 2025-05-08
Add `motion` and `emg` datatypes

JoeZiminski:


datashuttle/configs/canonical_configs.py
```diff
@@ -370,18 +402,35 @@ def in_place_update_settings_for_narrow_datatype(settings: dict):
...
+    # Copy the pre-existing settings for broad datatypes
     for key in ["behav", "ephys", "funcimg", "anat"]:
-        new_create_checkbox_configs[key]["on"] = settings["tui"][
+        new_create_checkbox_configs[key]["on"] = user_settings["tui"][
             "create_checkboxes_on"
         ][key]
-        new_transfer_checkbox_configs[key]["on"] = settings["tui"][
+        new_transfer_checkbox_configs[key]["on"] = user_settings["tui"][
             "transfer_checkboxes_on"
         ][key]
```
cs7-shrey, line 412: I am not sure but atm, this can break things now especially when we're adding more narrow datatypes. There are going to be cases which although use the newer version of datashuttle i.e `has_narrow_datatypes` is `True` but `is_not_missing_narrow_datatypes` is `False` (because of "motion" and "emg"), in that case, `user_settings["tui"]["transfer_checkboxes_on"][key]` is still a dict but we're assigning it to a boolean value, `new_transfer_checkbox_configs[key]["on"]`

So, I'm assuming the `has_narrow_datatypes` flag can provide us enough information about where `user_settings["tui"]["transfer_checkboxes_on"][key]` is a dict (newer datashuttle version ) or not (older datashuttle version). We can then do the assignment according to the flag.

datashuttle/configs/canonical_configs.py
```diff
@@ -370,18 +402,35 @@ def in_place_update_settings_for_narrow_datatype(settings: dict):
...
-    settings["tui"]["create_checkboxes_on"] = new_create_checkbox_configs
-    settings["tui"]["transfer_checkboxes_on"] = new_transfer_checkbox_configs
+    for narrow_datatype in all_narrow_datatypes:
+        if (
+            narrow_datatype
+            not in user_settings["tui"]["create_checkboxes_on"].keys()
+        ):
+            user_settings["tui"][
+                "create_checkboxes_on"
+            ] = new_create_checkbox_configs
```
cs7-shrey, line 427: If this part of the code is intended to copy the user_settings' narow datatypes config to the new_create_checkbox_configs, we should probably do it something like
```py
for narrow_datatype in all_narrow_datatypes:
    if (
        narrow_datatype
        in user_settings["tui"]["create_checkboxes_on"].keys()
    ):
        new_create_checkbox_configs[narrow_datatype] = user_settings["tui"][
            "create_checkboxes_on"
        ][narrow_datatype]
```
same we can do for `transfer_checkboxes_on` below and at the end we can do
```py
user_settings["tui"]["create_checkboxes_on"] = new_create_checkbox_configs
user_settings["tui"]["transfer_checkboxes_on"] = new_transfer_checkbox_configs
```
This will both maintain the order of key-value pairs in the settings and also ensure that previously set settings are respected.

datashuttle/configs/canonical_configs.py
JoeZiminski: Of course this only covers one of the two cases, thanks!

datashuttle/configs/canonical_configs.py
JoeZiminski: Agreed 👍

JoeZiminski:
Hey @cs7-shrey thanks a lot for this review and catching so many mistakes! Really goes to show the importance of code review, and writing tests (which I had lazily skipped 😅). I've rewritten the function as suggested and added some regression tests, let me know what you think.

tests/tests_regression/test_backwards_compatibility.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+class TestBackwardsCompatibility:
+
+    def test_v0_6_0(self):
+        """
+        v0.6.0 is the first version with narrow datatypes, and the checkboxes was refactored to
+        be a {"on": bool, "displayed": bool} dict rather than a bool indicating whether the checkbox is on.
+        However, this version is missing narrow datatypes added later (e.g. "motion").
+        In the test file, all 'displayed' are turned off except f2pe.
```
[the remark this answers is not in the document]
JoeZiminski: Hey @cs7-shrey this is a great idea and would ideally be the best approach. In practice, it is a little fiddly because the script would have to do something like uninstall datashuttle, reinstall the version and then build the `yaml` file settings depending on the `datashuttle` version. Because it's anticipated that we shouldn't have to do this too often, to avoid that complexity as you mention we can continue doing it manually for now. However, if we find we are continually having to manually create new `yaml` files, for basic stuff like adding a new narrow datatype, we should definitely take this approach. I'll resolve this for now but good to think about for the future.

cs7-shrey (approved):
LGTM. Just one small comment.

PR #528, opened 2025-06-20
[Prototyping] Play around with google drive and AWS testing with github secrets

JoeZiminski:
Hey @cs7-shrey this is PR to prototype a possible approach to testing. This PR will not be merged, it's just here as a rough example.

 I set up a NIU google drive and AWS with the credentials as github secrets (I can give you these credentials later you can play with this outside of the CI). The tests connect programmatically to google drive or AWS, setting up the connection and transferring some files. Later we can fill this in to perform some checks that the expected files were transferred. 
To have a look at the changes, you can check the diff on the last two commits (one for google drive and one for AWS).

For google drive, I had to change the way that the non-browser connection worked in order to make it programmatic. Now, the user must generate a credentials json through google, and pass the path of this to rclone. A benefit is that it does not require installing rclone on a second machine, but a downside is that it requires a bit more set up on the google end. For now, I'm leaning towards using this method because it can be performed programmatically, and it's roughly equivalent in convenience, and the non-browser case should in any case be infrequently used. Let me know what you think about this, thanks!

### DIFF
--- a/.github/workflows/code_test_and_deploy.yml
+++ b/.github/workflows/code_test_and_deploy.yml
@@ -57,14 +57,23 @@ jobs:
           pip install .[dev]
 #      - name: Test
  #       run: pytest
-      - name: Set up Google Drive secrets
-        run: |
-          printf '%s' '${{ secrets.GDRIVE_SERVICE_ACCOUNT_JSON }}' > "$HOME/gdrive.json"
-          echo "GDRIVE_SERVICE_ACCOUNT_FILE=$HOME/gdrive.json" >> $GITHUB_ENV
-          echo "GDRIVE_ROOT_FOLDER_ID=${{ secrets.GDRIVE_ROOT_FOLDER_ID }}" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


JoeZiminski:
Some general thoughts:
1) We could create a new branch /connection-dev (I'll make sure you have permissions to branch off the main repo) and squash merge #503 into it, and then start a new PR for tests. This might make it easier to track, because the implementation is basically done and it will be nice to draw a line under it and then start on tests. But, we cannot merge into main yet as I think there will be a quick release done before this is complete, and we can't release it without the tests.
2) The search  you have using `rclone` is really nice, if we can centralise local filesystem and ssh tests through this function, it would be fantastic. We could then perform all of the heavy-duty tests for local filesystem, and then do a few key tests for ssh, aws and gdrive, because they will all share most of the underlying code.
3) I'm working on the ssh tests, I can refactor these according to (2) assuming that it is possible. Then we will have the 'connections' implementation and test really tight!

PR #547, opened 2025-07-10
Add Python 3.13 support.

JoeZiminski:
Update `pyproject.toml` and CI tests to support Python 3.13.

Currently the conda-tests are not changed because a release needs to be made on the Python side to trigger a conda release. Then, a follow-up PR will add the conda release tests.


alessandrofelder (approved):
Happy to approve.

From a naive repo search for "3." you might want to double-check that you are not using the `tox.ini` (don't think you are) and the conda-install-check workflow (which I think you are using, and is missing python 3.13), and add 3.13 if needed?

PR #505, opened 2025-04-05
Correct `mainwindow` typehints

cs7-shrey:
## Description
This PR changes the typehint of `mainwindow` from `App` to `TuiApp` for better type checking and code completion in IDEs.

**What is this PR**

- [ ] Bug fix
- [ ] Addition of a new feature
- [x] Other

**Why is this PR needed?**
To correct typehints of the `mainwindow` parameter used in the codebase.

## How has this PR been tested?
Tested with pre-commit.

## Is this a breaking change?
No

## Does this PR require an update to the documentation?
No

## Checklist:

- [x] The code has been tested locally
- [ ] Tests have been added to cover all new functionality
- [ ] The documentation has been updated to reflect any changes
- [x] The code has been formatted with [pre-commit](https://pre-commit.com/)

PR #557, opened 2025-07-30
Keep default drive for SelectDirectoryTreeScreen as home.

### LINKED ISSUE #550: Set the default folder to the user home when setting up configs
Currently it defaults to the drive base, but in general it will be more user-friendly to default to the users home drive

JoeZiminski:
This PR closes #550 by ensuring that the default path on the DirectoryTree in the select path window (used for new project and validate screen) is the home path. The problem is that the drive select (e.g. select C:\, D:\) was automatically triggered (I think when it is composed) calling `on_select_changed` which reset the directory tree to the drive. 

I couldn't find a way around this except to use a flag, which isn't ideal but it works. Now the default home path is used and only reset to the drive base when the select is changed. Also, I removed the `path_` argument for `SelectDirectoryTree` because it was never used.

Docs do not need changing and a test is updated.

### DIFF
--- a/datashuttle/tui/screens/modal_dialogs.py
+++ b/datashuttle/tui/screens/modal_dialogs.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import asyncio
-from typing import TYPE_CHECKING, Callable, Optional
+from typing import TYPE_CHECKING, Callable
 
 if TYPE_CHECKING:
     from pathlib import Path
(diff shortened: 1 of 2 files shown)


cs7-shrey (approved):

cs7-shrey:
Hey @JoeZiminski, you mistakenly mentioned #505, in the PR body. Can you change it to #550 ?

PR #559, opened 2025-07-31
Only refresh data type checkbox on select datatype change.

### LINKED ISSUE #542: Fields in Create Folders reset after clicking Displayed Datatypes
I was creating new folders in an existing project and after entering the subject and session name I clicked on `Displayed Datatypes`. This reset both the `Subject(s)` and `Session(s)` field. Is this the intended behaviour? I would expect the content of those field be maintained after I exit out of the `Displayed Datatypes` dialogue.

JoeZiminski:
This PR closes #542 by refreshing only the datatype-checkboxes widget rather than the entire screen, so that content of other widgets is not deleted.

Previously the entire tab was recomposed after the datatype checkboxes were changed, meaning that content in all other widgets was lost. Now, only the datatype checkboxes are refreshed. As far as I can tell, to achieve this in textual, the widget has to be entirely removed and then re-initialized. 

This PR adds to the tests ensuring that content in non-datatype checkbox widgets is not deleted after opening the display datatype checkbox window. No documentation is needed.

### DIFF
--- a/datashuttle/tui/tabs/transfer.py
+++ b/datashuttle/tui/tabs/transfer.py
@@ -141,7 +141,7 @@ class TransferTab(TreeAndInputTab):
             ),
             # These are almost identical to create tab
             Label("Datatype(s)", id="transfer_datatype_label"),
-            self.get_datatypes_widget(),
+            self.get_datatypes_checkboxes_widget(),
             self.get_displayed_datatypes_button(),
         ]
 


JoeZiminski:
Hey @IgorTatarnikov thanks for raising the issue on this, is this working for you?

IgorTatarnikov (approved):
Looks good to me! :+1:

PR #569, opened 2025-08-04
Add rclone install check only if call fails.

### LINKED ISSUE #560: Throw error on suggest next sub / ses if rclone not found
since #551 rclone is required for suggesting new sub or ses, however its not checked as a dependency when this function is used and now always returns `sub-001` or `ses-001`. Need to look into this a little more (why is it not throwing at error?) but everything works well when rclone is not installed. 

Rclone check should not happen on datashuttle `__init__` because it is slow and lags startup.

JoeZiminski:
closes #560. Previously a check for `rclone` install was performed on datashuttle startup if the project did not exist. The assumption was the first time the program is used, if `rclone` is not installed the install can be reminded. However, this doesn't make much sense because new environments are often created, but this doesn't mean a new project will be made.

The problem with the `rclone` install check is that it spawns a subprocess to call `rclone` which is slow, so it cannot be arbitrarily called. The reason it is not always called at program start up is because it creates an annoying lag when opening a project.

The solution here is to only check the rclone install if the rclone call has actually failed. This means we can avoid the overhead of checking up-front, and only incur it if the `rclone` call has failed (in which case, a short delay is not an issue).

No docs are required for this, it would be ideal to test it but it would be quite complex in the CI (having to uninstall `rclone` in the test environment) and is not easily mocked because of the subprocess call. In this case no tests are added, it is not critical and will come up in general day-to-day test usage if it fails.

@cs7-shrey the only problem with this is it might cause issues for the `Popen` method of calling rclone? Maybe it is not possible to take this approach in that function?

### DIFF
--- a/datashuttle/utils/rclone.py
+++ b/datashuttle/utils/rclone.py
@@ -195,14 +195,13 @@ def log_rclone_config_output() -> None:
 def prompt_rclone_download_if_does_not_exist() -> None:
     """Check that rclone is installed."""
     if not check_rclone_with_default_call():
-
         newline = "" if "PYTEST_CURRENT_TEST" in os.environ else "\n"
 
         utils.log_and_raise_error(
             f"RClone installation not found. Install by entering "
             f"the following into your terminal:{newline}"
             f"  conda install -c conda-forge rclone",
-            RuntimeError
+            RuntimeError,
         )
 
 


datashuttle/utils/rclone.py
```diff
@@ -205,6 +202,15 @@ def prompt_rclone_download_if_does_not_exist() -> None:
         )
 
 
+def check_rclone_with_default_call() -> bool:
```
JoeZiminski, line 205: This change just swaps the order of the functions so that the called function is under the calling function

cs7-shrey:
Hey @JoeZiminski, this looks good. It is completely extendable to the Popen method (in the function that awaits the Popen call). However, I'm not sure of the performance benefits. When I ran it locally, the difference wasn't very noticeable. So, we might as well just keep the check that runs on DataShuttle `__init__` because it seems handy in some cases for ex. when creating a new project (assuming the user just installed datashuttle and is making their first project).

Checking rclone installed on failing commands is really good in general. I'm all in for that. However, #560 (the part where it returns sub-001 in case rclone is not installed), reflects a deeper problem and the "rclone not installed" is just one manifestation. So, checking if rclone installed or not would be more of a quick fix. Another way this problem manifests is, if you setup a project with SSH connection method and don't setup the connection (so that remote rclone calls fail), and then if you try to use suggest next sub or ses with `include_central`, it works just fine, essentially ignoring the remote. 

Here's why I it happens,
the `search_local_or_remote` function that was recently added for search remote using `rclone lsjson`, doesn't raise error on non zero returncode and returns empty list for files and folders. If we can raise error on a non zero return code it would solve this issue. But if it breaks things elsewhere, we can just leave it as it is.

JoeZiminski:
Thanks @cs7-shrey! There was an error in the previous implementation, the checking function called `call_rclone` again which led to infinite recursion, I wonder if this resulted in the error not being triggered when you tested. Would you mind trying again?  I tested in the scenarios you tried (nice idea BTW to set up SSH but don't connect, I will use that in future) and it raised the error where expected. (for the first subject, it creates sub-001 as default, then afterwards attempts to search using `rclone). I think in `search_local_or_remote` it _should_ raise the error within `call_rclone` so execution should be stopped before `output.returncode` is checked in the function body.

 On Windows, spawning a new process is [much slower than UNIX](https://stackoverflow.com/questions/47845/why-is-creating-a-new-process-more-expensive-on-windows-than-linux?utm_source=chatgpt.com), in part because UNIX systems use a 'fork' method by default in which the new process shares the memory space of the old process unless it writes to it, in which case it is copied ('copy-on-write'). On windows, the method spawns an entirely new process with dedicated memory etc. which is slower. On my slightly older laptop, unfortunately this lag is quite noticeable and makes the start-up process of datashuttle a little jarring. It's a shame because it creates all kinds of problems, including having to think about #571. I wish the world used UNIX systems!

cs7-shrey:
Hey @JoeZiminski, I might have missed out on trying that part somehow. I tried it now and it works fine. I agree on the unix/windows part, sorry i was a bit biased with my operating system. Also, what do you think about raising error in the `search_local_or_remote` in case `call_rclone` returns a non zero returncode or do you think it'd be better to handle in another PR?

cs7-shrey (approved):

PR #571, opened 2025-08-05
Partially revert lsjson

JoeZiminski:
This PR partially reverts #551 by using local filesystem searches when possible for speed improvement (avoiding process creation which is slow on Windows).

It essentially creates a duplicate function, `search_local_filesystem` that should have the same output as `search_central_via_connection` but does not call `rclone`. This is faster as a new process does not need to be made with `subprocess.run()`. This slowness is particularly noticeable during live-validation on the create-folders screen, it is so bad that is makes the user experience very poor. On Linux, it is less of an issue.

This solution is not at all ideal, we now have two functions doing the same thing. When all searches were done through `search_central_via_connection`, it meant that all local filesystem tests implicitly tested the behaviour ssh, aws, google drive tests. There are many more local filesystem transfer tests because these are much quicker to run. Therefore, this duplication is reducing test coverage. The solution is just to add a few more explicit tests to ssh, aws and google drive #577 and in this PR, perform a set of tests to check that the outputs of these functions are the same under many conditions.

I thought another potential solution in #551 is to create a process on `datashuttle` with POpen and hold it for the lifetime of the class, and call it when needed. However, as far as I can tell this is not how POpen works. It will also be a high burden and possible source of bugs to manage the lifetime of the process on the datashuttle class. So I think this is a better solution, although not ideal.

### DIFF
--- a/datashuttle/utils/folders.py
+++ b/datashuttle/utils/folders.py
@@ -660,7 +660,7 @@ def search_local_filesystem(
         elif item.is_file():
             all_filenames.append(to_append)
 
-    return all_folder_names, all_filenames
+    return sorted(all_folder_names), sorted(all_filenames)
 
 
 def search_central_via_connection(
--- a/datashuttle/utils/folders.py
+++ b/datashuttle/utils/folders.py
@@ -722,4 +722,4 @@ def search_central_via_connection(
         else:
             all_filenames.append(to_append)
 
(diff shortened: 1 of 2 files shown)


cs7-shrey (commented):
Looks good @JoeZiminski. Just a few minor suggestions.

cs7-shrey (approved):

PR #553, opened 2025-07-26
Revert: Google drive connection method to config token based approach

cs7-shrey:
## Description

**What is this PR**

- [ ] Bug fix
- [ ] Addition of a new feature
- [x] Other

**Why is this PR needed?**
Reverting back to config token based approach for google drive connections as google cloud had some policy changes

**What does this PR do?**
Include back the config token method of setting up google drive connections in absence of a browser.

## References

Please reference any existing issues/PRs that relate to this PR.

## How has this PR been tested?
Tested locally

## Is this a breaking change?
No

## Does this PR require an update to the documentation?
No

## Checklist:

- [x] The code has been tested locally
- [ ] Tests have been added to cover all new functionality
- [ ] The documentation has been updated to reflect any changes
- [x] The code has been formatted with [pre-commit](https://pre-commit.com/)

PR #572, opened 2025-08-05
[Prototyping] Play around with github secrets for gdrive aws CI tests

JoeZiminski:
Just doing some tests as not to pollute #570 with many random testing commits

### DIFF
--- a/.github/workflows/code_test_and_deploy.yml
+++ b/.github/workflows/code_test_and_deploy.yml
@@ -77,7 +77,7 @@ jobs:
           pytest tests/tests_transfers/ssh
       - name: All Other Tests
         run: |
-          pytest --ignore tests/tests_transfers/ssh
+          pytest -k "test_aws_transfer or test_gdrive_transfer"
 
   build_sdist_wheels:
     name: Build source distribution
--- a/datashuttle/utils/aws.py
+++ b/datashuttle/utils/aws.py
@@ -20,7 +20,7 @@ def check_if_aws_bucket_exists(cfg: Configs) -> bool:
 
     bucket_name = get_aws_bucket_name(cfg)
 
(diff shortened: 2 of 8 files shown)


cs7-shrey:
Hey @JoeZiminski, I think the google drive transfer tests are failing because all the runners are trying to simultaneously read and write from the same google drive folders and so that causes some mismatch. Will fix this by adding a random uuid to the central paths. Thanks a lot for this! Just the aws part needs fixing and I'll take it from there. Cheers!

Also, a quick suggestion, you can temporarily modify `code_test_and_deploy.yml` to run only the tests necessary, That will save some time.

JoeZiminski:
Hey @cs7-shrey this is working on AWS now, I had to make a change to the IAM permissions. Also, I had to remove `location_constraint` from the config, for some reason AWS didn't like it (even though it is set to the correct region). See [this commit](https://github.com/neuroinformatics-unit/datashuttle/pull/572/commits/f9dc543874992e68a955dc7b402702d7efdace26) for the changes that got it to work.

A couple of the commits from [this PR](https://github.com/neuroinformatics-unit/datashuttle/pull/553) that I added to handle the `central_path` being `None` are not on this branch but I guess that's okay because they will be in `add_gdrive_aws_remote`. It might be worth rebasing this branch onto `add_gdrive_aws_remote` anyway just to double-check this case is working. But, there is one case that was not handled there, that in the above commit has a `# TODO: check` next to it. The current solution (if `central_path` is `None` essentially ignore the bucket check) is not ideal, maybe there is a better way to handle this. 

Let me know if these changes work in the other branch!

cs7-shrey:
Thanks a lot for this @JoeZiminski! I'll merge this into the tests branch.

PR #573, opened 2025-08-06
Add exit button

### LINKED ISSUE #545: Clearer instructions for exiting TUI
Could the instructions for exiting the TUI be moved to the [Walkthrough datashuttle](https://github.com/neuroinformatics-unit/datashuttle/blob/main/docs/source/pages/get_started/getting-started.md) page? `Ctrl-C` may not be immediately obvious to people unfamiliar with the command line.

Is it also possible to add an `Exit` button to the main menu?

JoeZiminski:
This PR closes #545 by adding an 'Exit' button to the main window. I tried to use `CTRL+C` for copy directly, but forgot that in some cases this still quits the application on Windows. I think under certain conditions the `textual` event loop is in a strange / awaiting state and the CTRL+C is not caught, defaulting to windows behavior which is to exit the application (also encountered in #458).

As such, `CTRL+Q` is maintained for copy. However now, `CTRL+C` does not exit the application, but causes a popup with the following information:

```
      self.notify(
          "Use CTRL+Q to copy from Inputs and DirectoryTrees.\n"
          "Use ESC or the 'Exit' button to quit the application.\n"
          "CTRL+C can be used to copy after highlighting text with the mouse while pressing 'shift'.",
          timeout=6,
      )
```

This is because I think it might be annoying if you were trying to copy for the application to exit. It also lets the user know about a cool trick that if you select while holding shift, you can copy any text (found by @cs7-shrey).

### DIFF
--- a/docs/source/pages/user_guides/choose-a-terminal.md
+++ b/docs/source/pages/user_guides/choose-a-terminal.md
@@ -82,4 +82,4 @@ with rendering errors.
 Once you've chosen a terminal, get started with
 ``datashuttle`` with our [Getting Started Tutorial](getting-started-walkthrough).
 
-To quit ``datashuttle`` in the terminal, press `CTRL+C`.
+To quit ``datashuttle`` in the terminal, press `CTRL+Q`.
--- a/datashuttle/tui/app.py
+++ b/datashuttle/tui/app.py
@@ -12,8 +12,7 @@ import pyperclip
 import showinfm
 import yaml
 from textual.app import App, ComposeResult
-from textual.binding import Binding
-from textual.containers import Container, Horizontal
+from textual.containers import Container
 from textual.widgets import Button, Label
 
 from datashuttle.configs import canonical_folders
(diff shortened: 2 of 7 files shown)


docs/source/pages/user_guides/choose-a-terminal.md
```diff
@@ -82,4 +82,4 @@ with rendering errors.
 Once you've chosen a terminal, get started with
 ``datashuttle`` with our [Getting Started Tutorial](getting-started-walkthrough).
 
-To quit ``datashuttle`` in the terminal, press `CTRL+C`.
+To quit ``datashuttle`` in the terminal, press `CTRL+Q`.
```
IgorTatarnikov, line 85: Should this be `Esc`?

IgorTatarnikov (commented):
Looks great! Is it worth adding a warning when `Esc` is hit? I worry about incidental key pushes leading to an unexpected exit.

JoeZiminski:
Hey @IgorTatarnikov in the end I added the warning window, it was pretty quick! If you could take a quick look whenever you have time that would be great. Cheers!

PR #589, opened 2025-09-04
Skip AWS and GDRIVE tests when environment variables not set.

JoeZiminski:
Currently AWS or GDrive tests fail when run locally if environment variables required for set up are not set. Now, these tests are just skipped.


JoeZiminski:
Thanks @cs7-shrey I forgot about those! I did a small refactor to include these in the transfer tests. Even though it's a bit strange to have TUI tests outside of `tests_tui` and with transfers, it makes it less likely to forget that these config tests also use the connection set up machinery in future. On balance I think that is slightly preferable.

cs7-shrey (approved):
LGTM. I think we can remove the directories`tests_tui/aws` and `tests_tui/gdrive`.

PR #590, opened 2025-09-04
Rename setup_google_drive_connection -> setup_gdrive_connection.

JoeZiminski:
This PR renames `setup_google_drive_connection` to `setup_gdrive_connection`. The reason is that for all other API arguments it is refered to as `gdrive` rather than `google drive`, so this is gone for consistency. Alternatively, `gdrive` references could be changed to `google_drive`, but some of these arguments are already quite long. I think it `gdrive` is a pretty intuitive and easy-to-remember abbreviation. What do you think about this @cs7-shrey ?

### DIFF
--- a/datashuttle/tui/interface.py
+++ b/datashuttle/tui/interface.py
@@ -38,7 +38,7 @@ class Interface:
         self.name_templates: Dict = {}
         self.tui_settings: Dict = {}
 
-        self.google_drive_rclone_setup_process: subprocess.Popen | None = None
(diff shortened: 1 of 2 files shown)


cs7-shrey:
It's a nice thought. I think we can rename other such variables too in `interface.py`. Like, `google_drive_rclone_setup_process`, `terminate_google_drive_setup`, etc.

JoeZiminski:
Thanks @cs7-shrey good shout! Have updated

PR #594, opened 2025-09-08
Rename `name templates` to `validation templates`

JoeZiminski:
This PR renames 'Name templates' to 'Validation templates' (these are custom templates that can be validated against). Name templates is unfinformative as a name so it needs to be changed.

Unfortunately it is used in many places across the codebase, including in the configs. Therefore, a new entry into the config-loader is created to find "name_template" as an entry and convert it to "validation_template", for backwards compatability. The name is also updated in the TUI.

Tests and docs are also updated. Note SonarCloud is failing, but it is safe to ignore in this case.,

### DIFF
--- a/datashuttle/tui/screens/create_folder_settings.py
+++ b/datashuttle/tui/screens/create_folder_settings.py
@@ -89,8 +89,8 @@ class CreateFoldersSettingsScreen(ModalScreen):
         ses_on = not sub_on
 
         explanation = r"""
-        A 'Template' can be set check subject or session names are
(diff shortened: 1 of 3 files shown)


viktorpm (approved):
All looks good. I added a few general comments

JoeZiminski:
Thanks Viktor! Have made the changes as suggested. The SonoarCloud is a code duplication issue, but it is in tests and in this case easeier to read with the duplication / gets too confusing to address the duplication.

Issue #537, opened 2025-07-03
Can tick contradicting tickboxes, and error message is unclear


alessandrofelder:
**Describe the bug**
Error message unclear and I can tick both "all" and e.g. "funcimg" (which I think shouldn't be possible). If I then hit "transfer" I get an error, which I don't understand... which options does "these" refer to?

**Expected behaviour**
* If I tick e.g. "funcimg", "all" should get deselected.
* It should be clear what "these options" refer to in the error message

**Screenshots**

<img width="1388" alt="Image" src="https://github.com/user-attachments/assets/ae42be5d-a627-4b01-b036-e86c36e6d932" />

<img width="1388" alt="Image" src="https://github.com/user-attachments/assets/1b28b828-9230-4dda-b7eb-51ccbeb36bf3" />

**Computer used (please complete the following information):**

M2 Macbook

PR #595, opened 2025-09-08
Improve error message on bad datatype combination.

JoeZiminski:
addresses #537 (but does not fully close) by improving the error message. #537 highlights that the `all` and single datatypes can be selected in the custom transfer screen at the same time. The resulting error message is confusing.

@alessandrofelder do you think this (error message below) is clearer? I tried to change the checkboxes themselves such that only one of these options at a time can be selected. It's possible, but becomes a little complex in terms of the code and making clear to the user why boxes are automatically becoming checked / unchecked.I think an error alone is workable 🤔, if the message is very clear.

image

### DIFF
--- a/datashuttle/utils/data_transfer.py
+++ b/datashuttle/utils/data_transfer.py
@@ -397,8 +397,8 @@ class TransferData:
             [name in ["all", "all_datatype"] for name in self.datatype]
         ):
             utils.log_and_raise_error(
-                "'datatype' must only include 'all' "
-                "or 'all_datatype' if these options are used.",
+                "If either `all` or `all_datatype` is selected for datatype, "
+                "it must be the only selected option.",
                 ValueError,
             )
 


JoeZiminski:
Yes I was also surprised when I re-checked this PR and saw I had said it was difficult 😂 for posterity, the reason it is a bit fiddly is:

Currently the options are `all`, `all_datatype`, `all_non_datatype` and the actual datatypes. If `all` is switched we need to turn of any selected datatype as well as `all_datatype`, `all_non_datatype`. If `all_datatype` is selected with need to switch off `all` and any datatype. If `all_non_datatype` we need to switch of all only. If any datatype is selected with need to switch of `all` if it is selected. I am a little worried it might be confusing for users to understand the pattern behind the checkboxes being turned off.

This logic would be implemented in [this function](https://github.com/neuroinformatics-unit/datashuttle/blob/cafd40ed7f62107db066e2e0cb1b48c41ead6eee/datashuttle/tui/screens/datatypes.py#L257) with conditionals. It would not be too bad, maybe a little messy to iterative over the list of ids for the datatypes and switch the checkboxes and corresponding configs on / off.

TBH it would not be _too_ bad, I will keep the issue open and if this keeps coming up with can look into it more. Thanks @alessandrofelder!

PR #611, opened 2025-09-26
Fix syntax error.

JoeZiminski:
This PR fixes a syntax error on the tests workflow yaml file which somehow was not caught in automatic tests.


JoeZiminski:
Merging this as the tests are running, but failing due to issues from #596 that were not caught because the tests failed silently because of this syntax error. So will raise a follow up PR to fix these issues.

PR #634, opened 2025-11-19
Handle empty string case.

JoeZiminski:
This PR fixes an issue that occurs when an empty string is passed as a local or central path. The string was not being converted to path as the value can be a string, Path or `None` and only `if value` was checked, so the empty string case was being treated as `None`, then not being converted to Path, leading to an error elsewhere.


JoeZiminski:
cheers @niksirbi !

Issue #631, opened 2025-11-04
Windows tests, python 13, suddenly take over 2 hours to run


JoeZiminski:
Since the below commit (a pre-commit, which is very strange) windows tests have gone from taking 51 minutes (already too long) to over 2 hours. It's not clear why this is, the first thing to do will be to try reverting the pre-commit PR but I cannot see how that would affect the PR run time. 

Some general recommendations [here](https://chadgolden.com/blog/github-actions-hosted-windows-runners-slower-than-expected-ci-and-you) but it seems like something else is going on. It might be that performance of windows runners on python 13 have degraded generally, though I can't find any posts on this.

<img width="1318" height="1084" alt="Image" src="https://github.com/user-attachments/assets/9a291655-7444-48fe-9cd2-6802ab90fc69" />

JoeZiminski:
see #635

PR #635, opened 2025-11-19
Add durations to pytest in CI.

JoeZiminski:
Windows tests for python 3.13 are taking a very long time to run, #631. This PR prints the time taken for every test to dig further int othis.

The issue really started after [this PR](https://github.com/neuroinformatics-unit/datashuttle/pull/614/files) which is just a pre-commit auto update. Will try downgrading these.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,9 +89,6 @@ include-package-data = true
 include = ["datashuttle*"]
 exclude = ["tests*", "docs*"]
 
-[tool.pytest.ini_options]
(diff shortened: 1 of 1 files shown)


JoeZiminski:
Current hypothesis is that some dependency updated when the tests started taking longer, and it only affected python 3.9+ but not python 3.9

JoeZiminski:
Removing codecov did actually trim the tests by 5 minutes or so, but probably worth keeping anyway. Will raise an issue to summarise this PR and implement elsewhere. #642

PR #641, opened 2025-11-28
Pin fancylog.

JoeZiminski:
Further to #635 this PR pins fancylog to avoid the release in which package versions was logged by default, that included a subprocess call. This slowed down tests a lot. Tests are now running at their usual duration.

TODO: reverse CI changes before merging


adamltyson:
FYI I raised https://github.com/neuroinformatics-unit/fancylog/issues/86 so we can track fancylog supporting datashuttle properly across Python versions.

PR #653, opened 2025-12-18
Dynamically turn off transfer checkboxes.

### LINKED ISSUE #537: Can tick contradicting tickboxes, and error message is unclear

JoeZiminski:
This PR properly fixes #537 as it is an issue that has come up a few times. The problem was that you could click mutually exclusive options for the custom transfer window datatype selection checkboxes. for example, you could click 'all' and 'behav' and it would then throw an errror on transfer. Now, the check boxes properly respond such that if you click one of a mutually exclusive set of options, all other options will be deselected.

Tests have been added, no additional documentation required.

### DIFF
--- a/datashuttle/tui/custom_widgets.py
+++ b/datashuttle/tui/custom_widgets.py
@@ -16,6 +16,7 @@ if TYPE_CHECKING:
 
     from datashuttle.tui.app import TuiApp
     from datashuttle.tui.interface import Interface
+    from datashuttle.tui.screens.datatypes import BaseDatatypeCheckboxes
 
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


JoeZiminski:
Cheers @alessandrofelder!

PR #669, opened 2026-01-15
Default to "all" subjects or sessions for custom transfer if input left blank

### LINKED ISSUE #619: Do not allow blank `sub` or `ses` in the custom transfer TUI
Currently in the TUI you can leave the 'subject' or 'session' box blank and it will run the transfer, and transfer nothing. This is very confusing, you would expect it to use 'all' in case the subject or session box is blank. If it is blank, we could convert to 'all' or raise an error saying these need to be filled it with some possible options ('all', 'all_sub', etc)

ishan372or:
## Description

- [x] Bug fix
- [ ] Other

**Why is this PR needed?**
Currently, leaving the subject or session fields blank in the "Custom Transfer" TUI causes the transfer to fail silently (it attempts to transfer but selects nothing). This confusing behavior was identified in Issue #619. This PR ensures that blank inputs are interpreted as a wildcard ("all"), improving user experience.

**What does this PR do?**
I updated the logic in `datashuttle/tui/tabs/transfer.py` to check for blank inputs. Specifically, since the TUI returns `[""]`  for empty input fields, the code now explicitly catches this case and defaults `sub_names` and `ses_names` to `"all"`.

## References
fixes #619

Manual Verification:

1.Launched the TUI using datashuttle launch.

2.Navigated to the Transfer tab and selected Custom mode.

3.Left `Subject(s)` and `Session(s)` blank and clicked Transfer.

4.Verified in the logs and terminal that the system correctly defaulted to "all" and attempted the transfer, rather than failing silently.

## Is this a breaking change?

No. This change is backward-compatible. It only modifies the behavior when the subject or session fields are strictly empty in the TUI, which previously resulted in a failed operation anyway.

No. This is a fix to align the TUI behavior with expected user assumptions (blank = all).

## Checklist:

### DIFF
--- a/tests/tests_tui/test_tui_widgets_and_defaults.py
+++ b/tests/tests_tui/test_tui_widgets_and_defaults.py
@@ -1105,7 +1105,7 @@ class TestTuiWidgets(TuiBase):
 
     @pytest.mark.asyncio
     async def test_transfer_checkboxes_dynamic_on_off(
-            self, setup_project_paths
+        self, setup_project_paths
     ):
         """
         This tests that mutually exclusive checkbox options turn


ishan372or:
Thanks for the review @JoeZiminski, I agree on updating the placeholders to display `(default: "all")` for both subject and session inputs. And yes, I’m completely fine with you pushing the test directly to my fork. Please go ahead—an explanation afterward would be great so I can understand the approach and learn from it. I will make the changes as soon as possible.

JoeZiminski:
Thanks @ishan372or, yes that's been an ongoing issue, hopefully fixed with #674 but unrelated to this PR. Will update the test next week, then I think this will be good to go! thanks a lot

PR #673, opened 2026-01-28
Fix for new pandas version

JoeZiminski:
Pandas 3.0.0 changed the way null entries are represented in tables, breaking a test. This fixes here, for full information see below.

image (3)


JoeZiminski:
Merging, tests failing due to #670

PR #681, opened 2026-02-07
Fix: Update keyboard shortcut tooltips for macOS compatibility (Fixes #617)

### LINKED ISSUE #617: Macbook keyboard shortcuts
**Describe the bug**
There are some keyboard shortcuts suggested in datashuttle, one of them is CTRL+Q for to copy full filepath to clipboard. Though this combo in mac os quits the application.

**To Reproduce**
Run datashuttle on a macbook and press the keys ~~CTRL~~ CMD + Q.

**Expected behaviour**
It should not close the application and copying I expect to be performed by COMMAND + C for consistency with the rest of the mac world. 

**Computer used (please complete the following information):**
- OS: Mac OS Sequoia 15.6.1

AdityaGupta716:
## What is this PR
- [x] Bug fix

## Why is this PR needed?
Issue #617 reported that the CTRL+Q keyboard shortcut tooltip is problematic on macOS because CTRL+Q quits the application on that platform. This PR updates the tooltip to emphasize CTRL+C (with SHIFT) instead, which is the correct way to copy the highlighted file path to the clipboard.

## What does this PR do?
Updates keyboard shortcut tooltips in `datashuttle/tui/tooltips.py` to clarify that users should use CTRL+C with SHIFT to copy the filepath to the clipboard. This avoids the accidental application quit on macOS when users press CTRL+Q.

### DIFF
--- a/tests/tests_unit/test_links.py
+++ b/tests/tests_unit/test_links.py
@@ -3,11 +3,11 @@ import validators
 from datashuttle.configs import links
 
 
-def test_links():
+def test_links() -> None:
     """Test canonical links are working. Unfortunately Zulip links cannot
     be validated.
     """
     assert validators.url(links.get_docs_link())
     assert validators.url(links.get_github_link())
     assert validators.url(links.get_link_github_issues())
-    # assert validators.url(links.get_link_zulip()) Zulip links fail even when valid...
+    # assert validators.url(links.get_link_zulip()) Zulip links fail even when valid


JoeZiminski:
Thanks @AdityaGupta716,  we will need to change the key combination in the code as well. Ideally, we will find all instances of "CTRL+Q" / "ctrl+q", see where they are used, and manually test with CTR+W after updating them to ensure everything is working.

AdityaGupta716:
@JoeZiminski CTRL W wont work on mac cause it will actually delete the things we will try to copy  in line i think CTRL K or maybe also CTRL Y can work

Issue #532, opened 2025-06-26
Extend mypy to tests and add type hints to transfer


JoeZiminski:
Currently, type hints are not added to tests to avoid the maintenance burden, and mypy does not check the tests. In general, it would be nice to have type hints on tests, espsecially where they aid in reading. This is particularly important for some cases, see [here](https://github.com/neuroinformatics-unit/datashuttle/pull/208#discussion_r2166137973). 

In a new PR, we can start by:
1) extend mypy to cover tests 
2) start adding type hints where useful, [here](https://github.com/neuroinformatics-unit/datashuttle/pull/208#discussion_r2166137973) first.
3) extend developer docs to explain the situation

PR #684, opened 2026-02-08
ci: run mypy on tests and add initial typing for ssh and tui tests

ishan372or:
## Description

**What is this PR**

- [ ] Bug fix
- [x] Other

**Why is this PR needed?**
Tests were previously excluded from mypy to avoid maintenance burden.
To evaluate feasibility and improve readability where helpful, this PR enables mypy checking for tests and adds an initial, minimal set of type hints. This helps surface real typing issues early and establishes patterns for future incremental improvements.

**What does this PR do?**
Enables mypy checking for test files.
Adds minimal, targeted type hints to selected SSH test files.

## References
Related issue: #532 – Extend type hints to tests

Ran `pre-commit run mypy --all-files` locally

## Is this a breaking change?
No

No

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -83,9 +83,6 @@ requires = [
 build-backend = "setuptools.build_meta"
 
 [tool.mypy]
-exclude = [
-    "tests/"
-]
 
 [tool.setuptools]
 include-package-data = true
--- a/tests/tests_transfers/ssh/base_ssh.py
+++ b/tests/tests_transfers/ssh/base_ssh.py
@@ -4,6 +4,7 @@ import os
 import platform
 import subprocess
 from pathlib import Path
+from typing import Iterator
 
 import pytest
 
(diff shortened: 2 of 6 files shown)


JoeZiminski (changes requested):
Thanks @ishan372or! please see review, I think instead of using `Any` we can narrow this to `Datashuttle` class. If we type hint the output of a function, or a single input, we should probably type the entire function (all inputs and outputs).

JoeZiminski:
Hi @ishan372or thanks for this. After some consideration and discussion with the team, it was decided not to add type hints to the tests for now. This is because it is a lot of work to add to all tests, and adds maintenance burden going forward. Nonetheless thanks for this contribution, it will be useful for us to revisit in case we do go ahead with the work to add type hints to all types.

ishan372or:
@JoeZiminski , Thanks for the update and for considering the contribution. Glad it could be useful for future reference happy to revisit or help if you decide to move forward with it later.

PR #691, opened 2026-02-24
Handle textual v8, unpin textual in pyproject.toml

### LINKED ISSUE #688: handle textual version 8
See major release: https://github.com/Textualize/textual/releases/tag/v8.0.0. Should be an easy fix, then can unpin `pyproject.toml` as changed #687

AdityaGupta716:
Fixes https://github.com/neuroinformatics-unit/datashuttle/issues/688

Changes
Updated textual dependency in pyproject.toml from >=3.4.0,<8 to >=8.0.0, removing the temporary pin added in https://github.com/neuroinformatics-unit/datashuttle/pull/689
Fixed InvalidSelectValueError in configs_content.py by replacing select.value = False with select.clear(), which is the correct API in textual v8
Before
Selecting an existing project crashed with
`InvalidSelectValueError:

Screenshot 2026-02-25 at 3 15 40 AM

After
Existing projects load correctly with textual v8.
Screenshot 2026-02-25 at 3 09 23 AM

### DIFF
--- a/datashuttle/tui/shared/configs_content.py
+++ b/datashuttle/tui/shared/configs_content.py
@@ -744,7 +744,10 @@ class ConfigsContent(Container):
             if cfg_to_load["aws_region"] is None
             else cfg_to_load["aws_region"]
         )
-        select.value = value
+        if value is False:
+            select.clear()
+        else:
+            select.value = value
 
     def setup_widgets_to_display(
         self, connection_method: ConnectionMethods
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ dependencies = [
     "fancylog>=0.4.2",
     "simplejson",
     "pyperclip",
-    "textual>=3.4.0,<8",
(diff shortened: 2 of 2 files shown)


JoeZiminski (approved):
Thanks @AdityaGupta716! This works great, I tested it also on v3.4.0 and it works well for old versions, so I kept the pin >=3.4.0 so any textual version can be used, just in case. But most users will default to the latest version. Cheers!

AdityaGupta716:
hii @JoeZiminski,  thanks for the testing and confirmation,i  looked into the failing test and it seems to be pre-existing , not related to this PR. The test asserts "size changed" in the error message, but the CI's rclone version now produces a different error string. The same issue exists in test_tui_transfer.py too. should i fix it in this PR?

JoeZiminski:
Hi @AdityaGupta716 sorry for the late reply, thanks again for this. Indeed the test failure is unrelated, I will just merge this now and the tests should be fixed on main. Cheers!

PR #693, opened 2026-02-25
Add waiting screen to validate and centralise CentralWaitingScreen

### LINKED ISSUE #604: Add waiting screen to validate
When validating `include_central` can be slow, add a waiting screen if not using `local_filesystem`. similar to get next sub / ses (centralise this if possible).

AdityaGupta716:
### What does this PR do?
- Adds a waiting screen when validating with `include_central=True` and the connection method is not `local_filesystem` (i.e. `ssh`, `aws`, `gdrive`), where the connection can be slow
- Centralises the waiting screen logic by replacing `SearchingCentralForNextSubSesPopup` and `ValidatingCentralPopup` with a single reusable `CentralWaitingScreen(message: str)` class

### Why is this PR needed?
Fixes #604

### What is the approach?
- `CentralWaitingScreen` takes a `message` string, displays a `Label` + `LoadingIndicator`, consistent with the pattern established in #484 and #479
- Validation runs in a `@work(exclusive=True, thread=True)` thread, UI updates use `call_from_thread`, matching the existing sub/ses suggestion pattern

### Is this a breaking change?
No

### How has this PR been tested?
All 65 TUI tests pass locally.

### Checklist
- [x] Tests have been added to cover all new functionality
- [x] The code has been formatted with pre-commit

### DIFF
--- a/datashuttle/tui/css/tui_menu.tcss
+++ b/datashuttle/tui/css/tui_menu.tcss
@@ -554,7 +554,7 @@ ConfirmAndAwaitTransferPopup {
 
 /* Suggest next subject / session loading pop up --------------------------------------------------- */
 
-SearchingCentralForNextSubSesPopup {
+CentralWaitingScreen {
     align: center middle;
 }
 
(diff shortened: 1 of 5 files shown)


JoeZiminski (approved):
Thanks a lot @AdityaGupta716! This is working very nicely and is super useful, cheers

Issue #710, opened 2026-03-18
Fix duplicate fancylog entry in pyproject.tom


JoeZiminski:
See [here](https://github.com/neuroinformatics-unit/datashuttle/blob/v0.8.0/pyproject.toml), we just need to set the lower limit for python < 3.11 rather than set it once before the two conditionals.

JoeZiminski:
closed by #715

Issue #671, opened 2026-01-24
Security: Passwords Stored in Plaintext Memory


dhiraj-143r:
PROBLEM: SSH, AWS, and Google Drive passwords are stored as plaintext strings in Python memory, making them vulnerable to memory dumps, process inspection tools, error logging, and swap files.

CURRENT CODE (Vulnerable):

password = input("Enter SSH password: ")
Password stays in memory with no cleanup

PROPOSED FIX:
1) Use getpass for secure input:

2) Clear password from memory after use:

JoeZiminski:
addressed in #717 and updates elsewhere to the logging which redacts any password-containing variables from error messages. It's not expected that these would be logged in the first place, but just in case.

PR #712, opened 2026-03-24
Fix duplicate fancylog dep, bare excepts, password cleanup

knQzx:
went through the codebase and found a few things:

- fancylog>=0.4.2 on line 19 of pyproject.toml conflicts with the python-version-specific entries below — removed the duplicate (#710)
- bare `except:` in validation.py and `except BaseException` in setup_ssh.py — narrowed both to `except Exception`
- password stays in the TUI input widget after use — now cleared right after setup (#671)
- typo in ssh.py docstring: "remove" -> "remote"

I emailed hello@neuroinformatics.dev about GSoC 2026 (knqzx0@gmail.com)

### DIFF
--- a/datashuttle/tui/screens/setup_ssh.py
+++ b/datashuttle/tui/screens/setup_ssh.py
@@ -168,6 +168,8 @@ class SetupSshScreen(ModalScreen):
             password
         )
 
+        self.query_one("#setup_ssh_password_input").value = ""
+
         if success:
             message = (
                 f"Connection set up successfully.\n"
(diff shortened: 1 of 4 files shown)


JoeZiminski:
Hi @knQzx thanks for this, would you be able to split these into three separate PRs? it will be easier to review if the PRs are smaller and targetted. BTW, just so you know datashuttle is not taking part in GSoC this year.

knQzx:
splitting into 3 separate PRs as requested, thanks for the feedback!

PR #715, opened 2026-03-30
remove duplicate fancylog dependency

knQzx:
the generic `fancylog>=0.4.2` on line 19 conflicts with the python-version-specific entries below it — this just removes the duplicate so the correct version-pinned ones are used.


JoeZiminski (approved):
Thanks @knQzx !

PR #716, opened 2026-03-30
replace bare except clauses with except Exception

knQzx:
narrowed down a bare `except:` in validation.py and an `except BaseException:` in setup_ssh.py to `except Exception` so they don't accidentally swallow things like KeyboardInterrupt.


JoeZiminski (approved):
Thanks @knQzx!

PR #717, opened 2026-03-30
clear password from TUI input after use

knQzx:
the password was sticking around in the input widget after setup — now it gets cleared right after use. also fixed a small "remove" -> "remote" typo in ssh.py.

### DIFF
--- a/datashuttle/tui/screens/setup_ssh.py
+++ b/datashuttle/tui/screens/setup_ssh.py
@@ -168,6 +168,8 @@ class SetupSshScreen(ModalScreen):
             password
         )
 
+        self.query_one("#setup_ssh_password_input").value = ""
+
         if success:
             message = (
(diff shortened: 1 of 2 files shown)


JoeZiminski:
Thanks @knQzx! One concern is that the password will disappear for the user once they have clicked 'Ok' which might seem confusing. Can you test this in the TUI and see how it feels? It might be worth deleting this only before moving to the next page.

knQzx:
good point - moved the clear to after the if/else block so the password stays visible until the page actually transitions. also keeps it available for the [REDACTED] replacement on failure

JoeZiminski:
Hi @knQzx thanks again for this, that update works perfect. Will merge now cheers for the contribution!

JoeZiminski (approved):

PR #726, opened 2026-06-05
[joss] make links to source material explicitly links

sneakers-the-rat:
Part of: https://github.com/openjournals/joss-reviews/issues/9642

It looks like PDF readers will autodetect these as links, but good to explicitly declare them as links for styling and also for the JATS generation

### DIFF
--- a/joss-paper/paper.md
+++ b/joss-paper/paper.md
@@ -97,7 +97,7 @@ Currently, NeuroBlueprint is designed for experiments in which subjects go throu
 
 # Availability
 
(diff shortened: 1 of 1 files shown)


JoeZiminski:
thanks @sneakers-the-rat! Tests are failing as I think `joss-paper` needs rebasing on the new main, will do this for completion.

JoeZiminski (approved):

JoeZiminski (approved):

PR #731, opened 2026-06-16
Fix logging tests

JoeZiminski:
Tests were [failing](https://github.com/neuroinformatics-unit/datashuttle/actions/runs/27517320756) due to an issue in in which the logger was expected to not have been set up yet, but was persisting, possibly having been set up in another test. This has never been an issue before so not exactly sure what's happening, maybe the order in which the tests are run has changed if pytest in a new version.

The solution is to tear down the logger before and after the offending test.


PolarBean (approved):

JoeZiminski:
cheers @PolarBean!

PR #740, opened 2026-07-02
Fix custom transfer widgets persisting

JoeZiminski:
This PR fixes a bug in which the widgets for the cumstom transfer could persist after switching back to transfer entire project / top level folder, once the number of checkboxes had been customised. This is because fresh instances of the checkboxes are generated after customising them, but old versions of the widgets persisted in a list that was shown / hidden when the transfer type was changed.

This is fixed by separating these dynamic widgets into their own list which is properly updated. Tests are added.


niksirbi (approved):
LGTM

